PR #19, opened 2019-10-27
fix SBM20 conversion factor selection, support Si22G GM tube

ThomasWaldmann:
SBM20: the correct conversion factor was not set up due to a code bug

Si22G: was not dealt with at all in previous code.
       now correct dummy code is in place, but the conversion factor value
       needs to be fixed still, see XXX FIXME (issue #20).


ThomasWaldmann:
(the code in if and else branch was the same, this caused the bug)

ThomasWaldmann:
ping! :-)

can you merge soon please, so that I can continue working on stuff without causing merge conflicts?

rexfue:
merged #19

PR #26, opened 2019-11-03
fix indentation / spacing, remove tabs

ThomasWaldmann:
no code changes.

trying to be at least "locally consistent".

trivial, please merge quickly.


ThomasWaldmann:
BTW, is there some commonly accepted style guide for C nowadays?

I'm coming from the Python world and there basically everybody agrees on PEP8 (mostly), like e.g. using 4 spaces to indent and never ever use tabs.

ThomasWaldmann:
BTW, same here:

Maybe our editors can be configured to automatically expand tabs to (e.g.) 4 spaces on save, to avoid a mess of mixed spaces and tabs...

rexfue:
merged
I don't know, if there is any standard, i guess, there are many! 
But i agree, 4 spaces is OK. And open brace on line of statement, not on extra line.

rexfue:
Change my answer to '2 spaces'. That's what is in the file and we will keep this.

ThomasWaldmann:
OK, so let's use 2. \o/

PR #76, opened 2019-11-19
move hw docs into docs/hardware/, fixes #63

### LINKED ISSUE #63: move stuff into docs/hardware
since recently, there is a `hardware` subdir inside `docs`.

but some hw related files (heltec pinout / differences) are outside.

guess we should move them into `docs/hardware/`.

ThomasWaldmann:


rexfue:
Agreed.

PR #77, opened 2019-11-20
Esp32 isr compat

### LINKED ISSUE #74: fix ISRs
Read a comment that arduino's interrupts/nointerrupts()` is not implemented for esp32.

Assuming this is still the case, we need to do it like there (also note the IRAM_ATTR):

https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/Timer/RepeatTimer/RepeatTimer.ino

rexfue:
changed  isr routines to be esp32 compliant
deleted all unnecessary 'no interrupt()/interrupt()'
fixed a few global variables
moved attach_interrupt to end of setup()

fixes #74.


multigeiger/multigeiger.ino
```diff
@@ -223,23 +222,21 @@ volatile unsigned long isr_count_timestamp_2send= micros();
...
-
+         portMUX_TYPE  mux = portMUX_INITIALIZER_UNLOCKED;
 
 int Serial_Print_Mode = SERIAL_DEBUG;
 
 //====================================================================================================================================
 // ISRs
-void isr_GMC_capacitor_full() {
+void IRAM_ATTR isr_GMC_capacitor_full() {
   GMC_cap_full = 1;
```
ThomasWaldmann: please add a comment here about why acquiring/releasing the portmux is not needed here.

multigeiger/multigeiger.ino
```diff
@@ -561,17 +554,18 @@ void loop()
...
   }
 
-  if ((Serial_Print_Mode == Serial_Statistics_Log) && (counts_before != isr_GMC_counts)) {  // statistics log active?
-    noInterrupts();
-    Serial.println(isr_count_time_between, DEC);
-    counts_before = isr_GMC_counts;
-    interrupts();
+  if ((Serial_Print_Mode == Serial_Statistics_Log) && (gotGMCpulse)) {   // statistics log active?
+    unsigned int count_time_between;
+    portEXIT_CRITICAL(&mux);
```
ThomasWaldmann: ENTER

ThomasWaldmann:
Ah, as suspected, my mini-PR from the workshop (removing one not needed global var) caused a conflict now, which must be resolved before it can be merged.

ThomasWaldmann:
@rexfue maybe try this:
```
git checkout master
git pull eco master  # update your local master branch from ecocurious master
git checkout esp32_isr_compat
git rebase -i master
```
The last command tries to put your changeset on top of the current master.
Maybe you can use that to get rid of the 3 strange "merge commits".

rexfue:
changes done

multigeiger/multigeiger.ino
ThomasWaldmann: see my comment in #74.

looks like we need the portENTER_CRITICAL_ISR(&mux) + ...EXIT... calls here.

also i suggest that we use 2 different MUXes:
- one for the GM pulse counter resource
- one for the HV capacitor full resource.

rexfue:
Good idea to use different mutexes.

rexfue:
updated

rexfue:
solved conflict and merged #77

PR #95, opened 2019-12-08
many calls to millis() consolidated 

### LINKED ISSUE #91: lots of millis() calls
there are quite a lot of them.

for many CPUs, a subroutine call is rather expensive, so guess we could do something like:

```
ms = millis();  // once
foo = ms;
...  // assuming this is all relatively quick stuff
bar = ms;
```

rexfue:
fixes #91 

left mills() in variable initialisation and in two subroutines


ThomasWaldmann:
looks good. merge?

PR #99, opened 2019-12-08
fix not working speaker and led tick, fixes #97

### LINKED ISSUE #97: speaker / led tick not working any more
with current master branch: no ticks, no flashing.

cpm display seems to work though.

ThomasWaldmann:


multigeiger/multigeiger.ino
```diff
@@ -439,7 +439,16 @@ void loop()
...
+  unsigned long current_ms = millis();                           // to save multiple calls to millis()
+
+  // copy values from ISR
+  portENTER_CRITICAL(&mux_GMC_count);                            // enter critical section
+  GMC_counts = isr_GMC_counts;
+  isr_GMC_counts = 0;
+  count_timestamp = isr_count_timestamp;
+  GMC_counts_2send = isr_GMC_counts_2send;
+  count_timestamp_2send = isr_count_timestamp_2send;
+  portEXIT_CRITICAL(&mux_GMC_count);                             // leave critical section
```
ThomasWaldmann: this needs to be done here so GMC_counts (and others) reflect fresh values.

later, we check for GMC_counts in multiple conditions and at the end of the loop, we set GMC_counts to 0 (which seems somehow superfluous?).

multigeiger/multigeiger.ino
```diff
@@ -488,16 +497,6 @@ void loop()
   // Check if there are enough pulses detected or if enough time has elapsed.
   // If yes, then it is time to charge the HV capacitor and calculate the pulse rate.
   if ((GMC_counts >= MAXCOUNTS) || ((current_ms - time2display) >= DISPLAYREFRESH )) {
-    // copy values from ISR
-    portENTER_CRITICAL(&mux_GMC_count);                               // enter critical section
-    GMC_counts = isr_GMC_counts;
```
ThomasWaldmann: this can't be done in this if-block as the if-condition already expects GMC_counts to be set and "fresh".

rexfue: OK, i see the problem. Shall i merge or will you do it?

ThomasWaldmann: merged :)

PR #100, opened 2019-12-08
LED/speaker tick: fix conditions for alternative delay

ThomasWaldmann:
if we want the speaker tick, the tick generation does a delay of 4ms total.
else we need to do an alternative delay of 4ms (so that the LED is on
for 4ms) - but ONLY if the LED is enabled.

this fixes 2 inconsistent conditions for speaker and LED.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -465,10 +465,10 @@ void loop()
         digitalWrite (PIN_SPEAKER_OUTPUT_N, LOW);
         delayMicroseconds(500);
       }
-    }
-
-    if(ledTick & !speakerTick) {
-      delay(4);
+    } else {
+      if(ledTick && sw[LED_ON]) {
+        delay(4);
+      }
     }
     if(ledTick && sw[LED_ON]) {
       digitalWrite(LED_BUILTIN, LOW);     // switch off LED


multigeiger/multigeiger.ino
```diff
@@ -465,10 +465,10 @@ void loop()
         digitalWrite (PIN_SPEAKER_OUTPUT_N, LOW);
         delayMicroseconds(500);
       }
-    }
-
-    if(ledTick & !speakerTick) {
```
ThomasWaldmann: this did not check the 2 switches.

rexfue:
Oha, beste Dank. Das hab ich leider übersehen.

multigeiger/multigeiger.ino
```diff
@@ -465,10 +465,10 @@ void loop()
...
-      delay(4);
+    } else {
```
ThomasWaldmann: this is now the exact opposite of the if-condition (as it should be), so includes the speaker switch check.

multigeiger/multigeiger.ino
```diff
@@ -465,10 +465,10 @@ void loop()
...
+      if(ledTick && sw[LED_ON]) {
```
ThomasWaldmann: this now includes the LED switch check and is consistent with the other 2 checks that turn the pin on/off.

PR #102, opened 2019-12-08
only reset isr_GMC_counts if the display will be updated, fixes #101

### LINKED ISSUE #101: cpm value is permanently 0
looks like while fixing #97, i introduced a new bug:

the cpm display value is always staying at 0.

ThomasWaldmann:
also: move some not so timing critical stuff to the end of the loop.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -440,18 +440,7 @@ void loop()
   unsigned int HV_pulse_count;
   char sw[4];
   unsigned long current_ms = millis();                           // to save multiple calls to millis()
-
-  // copy values from ISR
-  portENTER_CRITICAL(&mux_GMC_count);                            // enter critical section
-  GMC_counts = isr_GMC_counts;
-  isr_GMC_counts = 0;
-  count_timestamp = isr_count_timestamp;
-  GMC_counts_2send = isr_GMC_counts_2send;
-  count_timestamp_2send = isr_count_timestamp_2send;
-  portEXIT_CRITICAL(&mux_GMC_count);                             // leave critical section
-
-  // Loop for IoTWebConf
-  iotWebConf.doLoop();
+  bool update_display;
 
   // Read Switches (active LOW!)
   sw[0] = !digitalRead(PIN_SWI_0);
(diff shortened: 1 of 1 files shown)


multigeiger/multigeiger.ino
```diff
@@ -440,63 +440,30 @@ void loop()
...
-  }
-
   #define DISPLAYREFRESH 10000
   #define MAXCOUNTS 100
+
+  // copy values from ISR
+  portENTER_CRITICAL(&mux_GMC_count);                            // enter critical section
+  GMC_counts = isr_GMC_counts;
+  update_display = (GMC_counts >= MAXCOUNTS) || ((current_ms - time2display) >= DISPLAYREFRESH);
+  if(update_display) isr_GMC_counts = 0;
```
ThomasWaldmann: this is new.

we only want to reset isr_GMC_counts to 0, IF we update the display.
but we need to do this in a critical section.
thus, compute update_display here (quick) and re-use it in the following if-condition.

multigeiger/multigeiger.ino
```diff
@@ -552,7 +519,6 @@ void loop()
         lastMinuteLog       = current_ms;
       }
     }
-    GMC_counts = 0;  // initialize ISR values
```
ThomasWaldmann: removed this. it is not needed as the correct value is assigned at the beginning of the loop function and it never needs to be changed until the next loop() invocation.

also, the comment was wrong, this is not an isr value, but a copy of it.

ThomasWaldmann:
display works again, but the ticking is a bit too regular in my ears.

multigeiger/multigeiger.ino
```diff
@@ -440,50 +440,28 @@ void loop()
...
   unsigned long current_ms = millis();                           // to save multiple calls to millis()
-
-  // copy values from ISR
-  portENTER_CRITICAL(&mux_GMC_count);                            // enter critical section
-  GMC_counts = isr_GMC_counts;
-  isr_GMC_counts = 0;
-  count_timestamp = isr_count_timestamp;
-  GMC_counts_2send = isr_GMC_counts_2send;
-  count_timestamp_2send = isr_count_timestamp_2send;
-  portEXIT_CRITICAL(&mux_GMC_count);                             // leave critical section
```
ThomasWaldmann: moved this closer to the if-update_display

multigeiger/multigeiger.ino
```diff
@@ -440,50 +440,28 @@ void loop()
...
-
-  // Loop for IoTWebConf
-  iotWebConf.doLoop();
```
ThomasWaldmann: this is not timing critical, moved to end of loop.

multigeiger/multigeiger.ino
```diff
@@ -440,50 +440,28 @@ void loop()
...
+  #define DISPLAYREFRESH 10000
+  #define MAXCOUNTS 100
 
-    if(speakerTick && sw[SPEAKER_ON]) {   // make "Tick" sound
-      for (int speaker_count = 0; speaker_count <= 3; speaker_count++) {
-        digitalWrite (PIN_SPEAKER_OUTPUT_P, LOW);
-        digitalWrite (PIN_SPEAKER_OUTPUT_N, HIGH);
-        delayMicroseconds(500);
-        digitalWrite (PIN_SPEAKER_OUTPUT_P, HIGH);
-        digitalWrite (PIN_SPEAKER_OUTPUT_N, LOW);
```
ThomasWaldmann: moved to end of loop.

multigeiger/multigeiger.ino
```diff
@@ -440,50 +440,28 @@ void loop()
...
-    } else {
-      if(ledTick && sw[LED_ON]) {
-        delay(4);
-      }
-    }
-    if(ledTick && sw[LED_ON]) {
-      digitalWrite(LED_BUILTIN, LOW);     // switch off LED
-    }
-    last_GMC_counts = GMC_counts;         // notice old value
-  }
```
ThomasWaldmann: moved to end of loop.

multigeiger/multigeiger.ino
```diff
@@ -440,50 +440,28 @@ void loop()
...
+  // copy values from ISR
+  portENTER_CRITICAL(&mux_GMC_count);                            // enter critical section
+  GMC_counts = isr_GMC_counts;
+  // Check if there are enough pulses detected or if enough time has elapsed.
+  // If yes, then it is time to calculate the pulse rate, update the display and recharge the HV capacitor.
+  update_display = (GMC_counts >= MAXCOUNTS) || ((current_ms - time2display) >= DISPLAYREFRESH);
+  if(update_display) isr_GMC_counts = 0;
```
ThomasWaldmann: this is new.

we only want to reset isr_GMC_counts to 0, IF we update the display (which is slow).
but we need to do this in a critical section.

thus, compute update_display here (quick) and re-use it in the following if-condition.

multigeiger/multigeiger.ino
```diff
@@ -440,68 +440,40 @@ void loop()
...
-  #define MAXCOUNTS 100
-  // Check if there are enough pulses detected or if enough time has elapsed.
-  // If yes, then it is time to charge the HV capacitor and calculate the pulse rate.
-  if ((GMC_counts >= MAXCOUNTS) || ((current_ms - time2display) >= DISPLAYREFRESH )) {
+  if(update_display) {
     time2display = current_ms;
     time_difference = count_timestamp - last_count_timestamp; // calculate all derived values
     last_count_timestamp = count_timestamp;                   // notice the old timestamp
-    HV_pulse_count = jb_HV_gen_charge__chargepulses();        // charge HV capacitor
-    hvpulsecnt2send += HV_pulse_count;                        // count for sending
```
ThomasWaldmann: we only want to recharge hv at one place in the loop.

avoids double recharging with no time in between.

multigeiger/multigeiger.ino
```diff
@@ -440,68 +440,40 @@ void loop()
...
 
-  // Pulse the high voltage at least every second
-  #define HVPULSERATE 1000
-  if((current_ms - time2hvpulse) >= HVPULSERATE ) {
+  // Pulse the high voltage if we got enough GMC pulses to update the display or at least every 1000ms.
+  #define HVPULSE_MS 1000
+  if(update_display || (current_ms - time2hvpulse) >= HVPULSE_MS ) {
     HV_pulse_count = jb_HV_gen_charge__chargepulses();      // charge HV capacitor - restarts time2hvpulse!
     hvpulsecnt2send += HV_pulse_count;                      // count for sending
   }
```
ThomasWaldmann: small cleanup and fixed naming.

multigeiger/multigeiger.ino
ThomasWaldmann: (this is the main fix)

ThomasWaldmann:
guess this is ready to merge, please review.

PR #104, opened 2019-12-16
Wireless Stick (LoRa) display fixed

rexfue:
- fixed problem with SOFTWARE VERSION on LoRa code. Its's now 2 bytes long.
- change defaults in userdefines-example 
- added defines in platform-example.ini

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -688,12 +688,12 @@ void DisplayStartscreen(void){
   u8x8.drawString(0, 2, "Geiger-");
   u8x8.drawString(0, 3, " Counter");
   char rv[9];
-  strncpy(rv, revString, 4);
-  rv[4] = '\0';
+  strncpy(rv, revString, 5);
+  rv[5] = '\0';
   u8x8.drawString(2, 4, rv);
-  strncpy(rv, &revString[5], 4);
-  strncpy(&rv[4], &revString[10], 2);
-  strncpy(&rv[6], &revString[13], 2);
+  strncpy(rv, &revString[6], 4);
+  strncpy(&rv[4], &revString[11], 2);
+  strncpy(&rv[6], &revString[14], 2);
   rv[8] = '\0';
   u8x8.drawString(0, 5, rv);
 #else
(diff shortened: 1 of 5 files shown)


multigeiger/userdefines-example.h
```diff
@@ -5,14 +5,16 @@
 // and then edit userdefines.h as appropriate for your use case / hardware.
 //====================================================================================================================================
 
+#ifndef CPU
```
ThomasWaldmann: please add a comment that when using platformio, this is defined in platformio.ini.

multigeiger/userdefines-example.h
```diff
@@ -21,6 +23,10 @@
 //#define   SERIAL_DEBUG Serial_One_Minute_Log  // One Minute logging
 //#define   SERIAL_DEBUG Serial_Statistics_Log  // Lists time [us] between two events
 
+// ** Special debug info: 
+//if set to 1, print debug info on serial (USB) interface while sending to servers (madavi or luftdaten)
```
ThomasWaldmann: add a blank after the comment markup //.

multigeiger/multigeiger.ino
```diff
@@ -688,12 +688,12 @@ void DisplayStartscreen(void){
...
-  rv[4] = '\0';
+  strncpy(rv, revString, 5);
+  rv[5] = '\0';
   u8x8.drawString(2, 4, rv);
-  strncpy(rv, &revString[5], 4);
-  strncpy(&rv[4], &revString[10], 2);
-  strncpy(&rv[6], &revString[13], 2);
+  strncpy(rv, &revString[6], 4);
+  strncpy(&rv[4], &revString[11], 2);
+  strncpy(&rv[6], &revString[14], 2);
```
ThomasWaldmann: this is rather ugly and maybe should be solved in another way.

guess as it broke now (1.9 -> 1.10), it will break again (1.xx -> 2.0),

multigeiger/version.h
```diff
@@ -1,11 +1,16 @@
 // Revision history:
 
-const char* revString = "V1.10_2019-12-13";
-#define SOFTWARE_VERSION 1A
+const char* revString = "V1.11_2019-12-16";
```
ThomasWaldmann: the date is not really needed here, people can look it up in the changelog, if interested.

also, we could get rid of the "V" and just use "1.11". would also make tagging in git easier and checking out some specific tag.

ThomasWaldmann: when changing like i suggested, we do not need revString at all. it could be just computed from SOFTWARE_VERSION.

multigeiger/multigeiger.ino
rexfue: You're right, that's ugly. Will be changed in one of the next releases

multigeiger/userdefines-example.h
rexfue: ok

multigeiger/userdefines-example.h
rexfue: done

multigeiger/version.h
rexfue: We want to show the date on the start screen, so we need it.
I will separate version and date and calculate the display string from that.

ThomasWaldmann: I just add SOFTWARE_DATE to already existing SOFTWARE_VERSION and compute all from these.

ThomasWaldmann: Or just add SOFTWARE_DATE and compute all from SOFTWARE_*.

PR #108, opened 2019-12-19
changed definitions for version and version date

rexfue:
new calculations to display version on both different displays


multigeiger/multigeiger.ino
```diff
@@ -354,6 +355,27 @@ void setup()
     while (!Serial) {};
   }
 
+  // build revString
+  strcat(revString,software_version);
+  strcat(revString," ");
+  strcat(revString,software_version_date);
```
ThomasWaldmann: how about using `sprintf`?

you could then use "V%s %s" as format string and also remove the "V" initialiser in line 233.

this also removes the theoretical problem that the code in this PR only works once (which is no real problem, as setup() is only executed once, but still a bit unpretty.

multigeiger/version.h
```diff
@@ -1,9 +1,8 @@
-// Revision history:
+const char* software_version = "1.11";
```
ThomasWaldmann: Is it only x.y because we have no space on the small display?

A x.y.z scheme would be better so we can follow semantic versioning (== use .z for maintenance bug fixes made after initial x.y.0 release).

ThomasWaldmann: I would rather remove the date from display and have .z displayed.

multigeiger/multigeiger.ino
```diff
@@ -354,6 +355,27 @@ void setup()
...
+  char *p = (char *)software_version;
+  while (*p != '\0') {
+    if (*p != '.') {
+      loravers[i++] = *p;
+    }
+    p++;
+  }
+  loravers[i] = '\0';
+  lora_software_version[0] = atoi(loravers)/100;
+  lora_software_version[1] = atoi(loravers)%100;
```
ThomasWaldmann: that only works correctly for versions like:
- 1.00
- 1.01
- 1.99

but not for:
- 1.1
- 1.2
- 1.3.1 (in case we would use x.y.z)

multigeiger/multigeiger.ino
```diff
@@ -917,8 +948,8 @@ void sendData2TTN(int sendwhat, unsigned int hvpulses) {
   // next byte is the tube version
   ttnData[4] = tubes[TUBE_TYPE].nbr;
     // and last is software version
-  ttnData[5] = SOFTWARE_VERSION>>8;
-  ttnData[6] = SOFTWARE_VERSION&0xFF;
+  ttnData[5] = (unsigned char)lora_software_version[0];
+  ttnData[6] = (unsigned char)lora_software_version[1];
```
ThomasWaldmann: how about typing lora_software_version variable as `uint8 ... []` (or unsigned char, if uint8 does not exist)?

multigeiger/multigeiger.ino
rexfue: ok, sounds better

multigeiger/version.h
rexfue: The reason for not to have z is the extra byte for lora - if we have z, we must send 3 bytes. And in lora it is important so send as few bytes as possible.
And I think, the date should be on the display. We should discuss this with Jürgen @WwwBoehriDe and Nicola @ecocurious.

multigeiger/multigeiger.ino
rexfue: You are right. To make it easy I wanted to use always 2 digits for the lower number.

multigeiger/multigeiger.ino
rexfue: Could be done. Then we should cast the atoi to unsigned char.

multigeiger/multigeiger.ino
ThomasWaldmann: Then better put a comment into the source if you don't want to break the code.

It's a bit unusual though, normally 1.1 or 1.1.1 should be valid as a version number.

multigeiger/version.h
ThomasWaldmann: not necessarily.

1.2.3 could be encoded as 0x1023 - assuming that we will neved need major version > 15 or patch version > 15.

Not many patch versions can be ensured by releasing often.

Major version change means "big change" or even "breaking change" in semantic versioning - maybe we never ever need that.

ThomasWaldmann: (just an example with nice hex representation - we could use also octal 0oXYYZZ (== 15 bits))

multigeiger/multigeiger.ino
rexfue: Sorry, I guess i clicked the wrong button.

Nice idea, to use xyyz in one 16 bit variable. I agree, we never would use more than 15 big versions and more than 15 patches either. So I will change the versioning system in one of the next PRs.
Thanks for this solution.

PR #111, opened 2019-12-22
Rexfue rev string on minidisplay

rexfue:


ThomasWaldmann:
hmm, why did you make a new PR instead of pushing to the existing one?

ThomasWaldmann:
hmm, seems like this is based on the "as is" acceptance / merging of the other one.

so it might be the case that you can improve the other one without getting into the way of this one.

rexfue:
I'm not sure, what the problem was. I hope, i didn't mess it up too much. I couldn't push  my changes. Then I pulled a fresh version and tried to push. And GitHub wanted a new PR. 
I'm still on "Kriegsfuß" with this ;-)
Let's merge, and then all problems are gone - hopefully.
Other question: what is your environment on linux to compile this? I tried platformio on a PI and it worked.

ThomasWaldmann:
I use Debian Linux, git on the commandline and the Arduino IDE for compiling.

rexfue:
OK, I don't like Arduino-IDE because of the ugly editor. On MAC I use Visual Code with Platformio.

Do You agree to merge this?

multigeiger/multigeiger.ino
```diff
@@ -355,33 +355,27 @@ void setup()
...
+// build LoRa software version
+char softvers[8];
+strcpy(softvers,software_version);
+char *p;
+p = strtok(softvers,".");
+lora_software_version[0] = (uint8_t)atoi(p);
+p = strtok(NULL,".");
+lora_software_version[1] = (uint8_t)atoi(p);
+#endif
+
```
ThomasWaldmann: is that indented differently than the other code? please only use spaces.

multigeiger/multigeiger.ino
```diff
@@ -713,24 +707,20 @@ void DisplayStartscreen(void){
...
   char rv[9];
-  char delim[] = "_-";
-  char *p; 
-  p = strtok((char *)rvStr,delim); 
-  strcpy(rv, p);
+  sprintf(rv,"V%s",software_version);
   u8x8.drawString(2, 4, rv);
   *rv = '\0';
-  p = strtok(NULL,delim);
+  strcpy(softversdat, software_version_date);
```
ThomasWaldmann: sometimes you leave a space after comma (which is good, better readable), but often not.

would be greatly improve readability if you always did.

ThomasWaldmann: and can we please keep the "e", so have "date", not "dat"?

multigeiger/userdefines-example.h
```diff
@@ -5,10 +5,12 @@
 // and then edit userdefines.h as appropriate for your use case / hardware.
 //====================================================================================================================================
 
-#ifndef CPU           // if using platformio, CPU is defined in platformio.ini
-// ** select (uncomment) exactly one of following CPU (board type) defines:
-#define CPU WIFI                                // Heltec WiFi Kit 32 (big display, WiFi-only)
-// #define CPU STICK                             // Heltec Wireless Stick (small display, WiFi+LoRa)
+// If you use platformio, CPU is defined in platformio.ini and
+// you don't need to change the definition of CPU
```
ThomasWaldmann: Please add a "." at the end of sentence (always).

multigeiger/userdefines-example.h
```diff
@@ -5,10 +5,12 @@
...
+// Arduino-IDE users *must* define CPU here and uncomment one and only one of following CPU-Defines
```
ThomasWaldmann: In English, dashes ("Bindestrich") is much more rare than in German.
Usually it is just a blank ("Leerzeichen") instead.

multigeiger/multigeiger.ino
```diff
@@ -917,8 +924,8 @@ void sendData2TTN(int sendwhat, unsigned int hvpulses) {
   // next byte is the tube version
   ttnData[4] = tubes[TUBE_TYPE].nbr;
     // and last is software version
```
ThomasWaldmann: not in new code, but there seems to be a bad indentation here.

multigeiger/version.h
```diff
@@ -1,15 +1,18 @@
-// Revision history:
+// To fit in 16bit for lora version number we have
+// some limits here: max. VERSION_MAJOR = 15, max. VERSION_PATCH = 15
+#define VERSION_MAJOR 1
+#define VERSION_MINOR 11
+#define VERSION_PATCH 0
+#define VERSION_DATE "2019-12-16"
```
ThomasWaldmann: We could have a comment above this that this is a date in YYYY-MM-DD format.

It is clear with the current value, but not for every value.

ThomasWaldmann:
OK, guess that is all for now. Review finished.

ThomasWaldmann:
BTW, I do not use the arduino editor.

I switched the IDE to "external editor", use vim and only use the IDE to compile / flash.

multigeiger/version.h
```diff
@@ -1,8 +1,9 @@
 // To fit in 16bit for lora version number we have
-// some limits here: max. VERSION_MAJOR = 15, max. VERSION_PATCH = 15
+// some limits here: max. VERSION_MAJOR = 15, max. VERSION_PATCH = 15.
 #define VERSION_MAJOR 1
 #define VERSION_MINOR 11
 #define VERSION_PATCH 0
+// Date is in format "YYY-MM-DD".
```
ThomasWaldmann: one Y more please.

rexfue: done

PR #113, opened 2019-12-25
Sensorname in readme

rexfue:


ThomasWaldmann:
https://github.com/ecocurious/MultiGeiger/blob/master/multigeiger/multigeiger.ino#L180

https://github.com/ecocurious/MultiGeiger/blob/master/multigeiger/userdefines-example.h#L52

The source / config still uses "LUFTDATEN" as the name. 

sensor.community is a much better name IMHO, so change it everywhere?

Maybe just do a global search for "luft"?

PR #131, opened 2020-01-11
output error msg on Serial if HV charging fails

ThomasWaldmann:
also:
- add a constant MAX_CHARGE_PULSES
- lowered MAX_CHARGE_PULSES from 1000 to 333 (so it is << 1s as expected by main loop recharge)
- cosmetic changes (3rd commit)

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -138,9 +138,9 @@ enum {SEND_CPM,SEND_BME};
 #define MEASUREMENT_INTERVAL 150
 
 // How many HV capacitor charge pulses to generate before giving up.
-// (MAX_CHARGEPULSES * pulse_duration) should be less than the interval in
+// (MAX_CHARGE_PULSES * pulse_duration) should be less than the interval in
 // the main loop (currently 1000ms) where it unconditionally recharges.
-#define MAX_CHARGEPULSES 333
+#define MAX_CHARGE_PULSES 333
 
 // MAX time to wait until connected. [msec]
 // If there is still no connection after that time,
@@ -277,7 +277,7 @@ void IRAM_ATTR isr_GMC_count() {
 
 //====================================================================================================================================
 // Function Prototypes
-int jb_HV_gen_charge__chargepulses(int max_chargepulses);
+int gen_charge_pulses(int max_charge_pulses);
 void DisplayGMC(int TimeSec, int RadNSvph, int CPS);
 void SoundStartsound();
 void jbTone(unsigned int frequency_mHz, unsigned int time_ms, unsigned char volume);
@@ -448,7 +448,7 @@ void setup()
   attachInterrupt (digitalPinToInterrupt (PIN_GMC_count_INPUT), isr_GMC_count, FALLING);            // GMC pulse detected
   
   // charge hv capacitor
-  jb_HV_gen_charge__chargepulses(MAX_CHARGEPULSES);
+  gen_charge_pulses(MAX_CHARGE_PULSES);
 
 }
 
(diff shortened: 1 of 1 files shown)


multigeiger/multigeiger.ino
```diff
@@ -137,6 +137,9 @@ enum {SEND_CPM,SEND_BME};
 // Measurement interval (default 2.5min) [sec]
 #define MEASUREMENT_INTERVAL 150
 
+// How many HV capacitor charge pulses to generate before giving up:
+#define MAX_CHARGEPULSES 1000
```
ThomasWaldmann: btw, i noticed that with the current value of 1000, charging will take up to ~2.6s which is way too long considering we recharge in the main loop every 1000ms anyway, see:
```
  // Pulse the high voltage if we got enough GMC pulses to update the display or at least every 1000ms.
```

multigeiger/multigeiger.ino
```diff
@@ -681,8 +685,10 @@ int jb_HV_gen_charge__chargepulses() {
     digitalWrite(PIN_HV_FET_OUTPUT, LOW);               // turn the HV FET off
     delayMicroseconds(1000);
     chargepulses++;
-  } while ((chargepulses < 1000) && !isr_GMC_cap_full); // either a timeout or a capacitor full interrupt stops this loop
+  } while ((chargepulses < max_chargepulses) && !isr_GMC_cap_full); // either a timeout or a capacitor full interrupt stops this loop
   time2hvpulse = millis();                              // we just pulsed, so restart timer
+  if((chargepulses == max_chargepulses) && !isr_GMC_cap_full)
+    Serial.println("Error: HV charging failed!");       // pulsed a lot, but still the capacitor is not at desired voltage
```
ThomasWaldmann: just a quickfix here to not let the error be completely silent.

for users not having connected the serial terminal it of course would be better to also update some global "hv_good" state and display it on the oled display.

ThomasWaldmann:
Usually I see that it needs 1 pulse (from serial output). So lowering from 1000 to 333 should be no problem at all (and maybe could be lowered again to a way smaller value later).

ThomasWaldmann:
@WwwBoehriDe What do you think?

It's a new concept to me that functions have their creators initials in their name. :-)

WwwBoehriDe:
Hi,
for me it is no problem to change the name of the Function. It was just for me to separate between Arduino- and Selfmade functions. No worries!

The Problem I see is that the limit covers two cases:
- rechargeing that happens every second. For this the limit is OK.
- but it also is the limit for the first charging. When the Main-Capacitor is charged 0=>400 V. Here the pulses could vary (I did not further investigat) a lot over PVT. Probably here a Limit of 333 could be too low. And here we don't want to see the Warning in the Log. 
Probably its a good sollution to suppress warnings the first 5 charge-runs.

ThomasWaldmann:
i can supply more charge pulses for first call from setup() as the max pulses is a func param now.



On January 12, 2020 12:53:02 PM GMT+01:00, WwwBoehriDe <notifications@github.com> wrote:
>Hi,
>for me it is no problem to change the name of the Function. It was just
>for me to separate between Arduino- and Selfmade functions. No worries!
>
>The Problem I see is that the limit covers two cases:
>- rechargeing that happens every second. For this the limit is OK.
>- but it also is the limit for the first charging. When the
>Main-Capacitor is charged 0=>400 V. Here the pulses could vary (I did
>not further investigat) a lot over PVT. Probably here a Limit of 333
>could be too low. And here we don't want to see the Warning in the Log.
>
>Probably its a good sollution to suppress warnings the first 5
>charge-runs.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/ecocurious/MultiGeiger/pull/131#issuecomment-573407549

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ThomasWaldmann:
Merging now "as is", later more pulses from setup() call.

Issue #120, opened 2020-01-05
Firmware Updates


ThomasWaldmann:
### Context
- MultiGeiger likely sits in a sealed tube and the USB port can't be accessed easily without disassembling the device / breaking the seal / resealing it.
- We need to also consider users who are not very technical, so doing manual upgrades (no matter how) might be an issue.
- Guess we would like to keep the devices up-to-date with the latest stable compatible firmware version (for bug fixes and features).

### Upgrade options
Considering this, manual upgrades via USB cable or manual via WiFi using a browser are no good options.

What's left are automated upgrades via WiFi from a internet server.

### Ideas

If the update file is cryptographically signed by a release manager, its integrity and its authenticity can be verified after it is downloaded, before it is flashed (it could be even verified again after flashing [reading it from flash] before switching to it to verify the flash procedure).

That way, the security of the transmission channel or of the server does not matter at all (as long as the signing key[s] are kept secret). We would never flash a tampered firmware.

Also, as long as the device sometimes manages to correctly transfer the file, it would work ok. We would never flash a corrupted firmware (but just retry later).

## Priority

Rather high as it is a pain to upgrade the devices by other means (see "Context").

ThomasWaldmann:
## Implementation ideas

### ESP IDF based

Does this mix with Arduino stuff?

https://docs.espressif.com/projects/esp-idf/en/latest/security/secure-boot.html#signed-app-verification-without-hardware-secure-boot 

This pretty much is what we want, I guess.

### Arduino based

WwwBoehriDe:
There was the idea to seal the device completly using silica gel to reduce humidity. Currently it looks like if this is not necessary. So the current housing-design is extremly easy to open and not sealed at all. So you have access to the device and its Micro-USB-Connector in just some seconds.

Please check:
https://github.com/ecocurious/MultiGeiger/blob/master/docs/Bauanleitung/Aufbauanleitung_2019_13_09.pdf
Page 8

I think Reinhard has alredy looked into OTA and found out that there are different memory sizes available on the Heltec-PCBs, which is problematic for OTA. But anyway. Probably its worth to check it again.

ThomasWaldmann:
Even if you can reach the USB plug, it is still uncomfortable that way (and unlikely that all/most people frequently do that). Of course one does not need to install all the updates, but maintaining a lot of differently outdated or buggy versions likely is also not fun for the developers.

[...]

PR #137, opened 2020-01-11
Simple OTA

ThomasWaldmann:
based on ArduinoOTA.

### DIFF
--- a/multigeiger/version.h
+++ b/multigeiger/version.h
@@ -1,13 +1,16 @@
 // To fit in 16bit for lora version number we have
 // some limits here: max. VERSION_MAJOR = 15, max. VERSION_PATCH = 15.
 #define VERSION_MAJOR 1
-#define VERSION_MINOR 11
+#define VERSION_MINOR 12
 #define VERSION_PATCH 0
 // Date is in format "YYYY-MM-DD".
-#define VERSION_DATE "2019-12-16"
+#define VERSION_DATE "2020-01-12"
 
 /**** Revision history:
 * Version date        author
+*
+* V1.12 2020-01-12  tw (not released yet / WIP)
+*   - simple OTA (Over-The-Air) updates via web browser based upload
 *
 * V1.11 2019-12-16  rxf
 *   - defaults in userdefines-example.h changed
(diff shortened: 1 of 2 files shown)


multigeiger/version.h
```diff
@@ -1,14 +1,17 @@
 // To fit in 16bit for lora version number we have
 // some limits here: max. VERSION_MAJOR = 15, max. VERSION_PATCH = 15.
 #define VERSION_MAJOR 1
-#define VERSION_MINOR 11
+#define VERSION_MINOR 12
```
ThomasWaldmann: I incremented minor, so we can say that "OTA is present since 1.12.0".

multigeiger/version.h
```diff
@@ -1,14 +1,17 @@
...
 #define VERSION_PATCH 0
 // Date is in format "YYYY-MM-DD".
-#define VERSION_DATE "2019-12-16"
+#define VERSION_DATE "2020-01-12"
```
ThomasWaldmann: Funny, version 1.12 and today is 2020-01-12.

So, release today? :-)

ThomasWaldmann:
There is one fundamental issue with it yet, that was already present in the example code I used:

When updating via e.g. a laptop, the laptop's wifi is usually connected to the esp32 access point (**not** to the internet).

So, in the usual case that the laptop does not have another internet connection (e.g. via LAN or LTE), it will be offline.

But the updater html code wants to load jquery.js from the internet...

That does not work if one is offline, of course, but there is a workaround to first load the jquery.js into the cache, then it'll work even while being offline.

I'm not fluent in javascript, so I currently do not feel like rewriting the jquery based functionality in pure javascript (so it does not have to load anything from the internet), so I suggest living with the workaround for now and doing that later.

ThomasWaldmann:
About space usage:

Sketch uses 828662 bytes (42%) of program storage space. Maximum is 1966080 bytes.

So I don't see any space issue. In case we ever have one, we'll care for it in the future.

ThomasWaldmann:
To generate a binary for upload: `Sketch -> Export compiled binary` in the Arduino IDE.

ThomasWaldmann:
@rexfue adding that "IOTWebConf also has OTA" to #120 would have maybe saved me quite some work as I obviously missed that one while searching for a sane Updater.

But I'll check that one now...

PR #146, opened 2020-01-12
Use IotWebConf as a library

### LINKED ISSUE #143: duplication of IotWebConf into our codebase
Seems like we have a full copy of that.

Problems:
- the lib version is unknown (at least I could not find it anywhere)
- hard to update
- it is locally modified to add setThingName functionality

Maybe the modification could be done in a cleaner way by subclassing and extending that way?
If so, we could remove the lib code from git and just install it via arduino library manager, as usual.

ThomasWaldmann:
fixes #143


ThomasWaldmann:
any comment?

i tried it, seems to work.

rexfue:
Yes it work. But the problem with the hard coded names for Geiger-APi etc. remains, that you changed.
I have no idea, how to solve this; beside the same hack.

ThomasWaldmann:
What do you mean precisely?

"problem with the hard coded names for Geiger-APi" ?

rexfue:
This:
  this->_thingNameParameter = IotWebConfParameter("Geiger accesspoint SSID", "iwcThingName", this->_thingName, IOTWEBCONF_WORD_LEN);
  this->_apPasswordParameter = IotWebConfParameter("Geiger accesspoint password", "iwcApPassword", this->_apPassword, IOTWEBCONF_WORD_LEN, "password");
  this->_wifiSsidParameter = IotWebConfParameter("WiFi client SSID", "iwcWifiSsid", this->_wifiSsid, IOTWEBCONF_WORD_LEN);
  this->_wifiPasswordParameter = IotWebConfParameter("WiFi client password", "iwcWifiPassword", this->_wifiPassword, IOTWEBCONF_WORD_LEN, "password");

Its hard coded in IotWebConf.cpp.

ThomasWaldmann:
oh crap, i did not remember that hack... - i'll have a look ...

but in general, a lib should be used as a lib. otherwise maintaining / updating is just too much pain.

rexfue:
agreed!  Maybe we should write similar stuff by ourselves - but why reinvent the wheel?
The luftdaten people did it by themselves, so  may be  we can steel  steel some code :-)

ThomasWaldmann:
https://github.com/prampec/IotWebConf/issues/98

ThomasWaldmann:
@rexfue solved with the help of @prampec!

ThomasWaldmann:
merge?

rexfue:
OK

Von meinem iPad gesendet

PR #152, opened 2020-01-15
Better way to set thingName

rexfue:
I found IMHO a better way to set the default thingName


multigeiger/multigeiger.ino
```diff
@@ -304,6 +304,7 @@ void clearDisplayLine(int line);
 void handleRoot(void);
 void configSaved(void);
 char * nullFill(int n, int digits);
+char* buildSSID();
```
ThomasWaldmann: can we try to keep style consistent (see line directly above)?

rexfue: OK. Seems 'nullfill' is the only place, where there is the additional spaces.

PR #160, opened 2020-01-17
tag log output with "GEIGER: "

### LINKED ISSUE #85: log output gets mixed up
if one configures Serial_Logging or One_Minute_Log to output some tabular data, some other output gets in between the table rows:

- wifi init output from esp
- other output from multigeiger code

this should be either cleaned up (suppress other output) or tagged (so that one can filter the output one wants to see, e.g. with grep or other means).

rexfue:
remove "F(" at literal Strings  -> ESP32 puts const char * and literal strings in flash by default
changed log strings to "const char*"

fixes #85


multigeiger/multigeiger.ino
```diff
@@ -579,7 +580,7 @@ void loop()
     count_time_between = isr_count_time_between;
     isr_gotGMCpulse = 0;
     portEXIT_CRITICAL(&mux_GMC_count);
-    Serial.println(count_time_between, DEC);
+    Serial.printf("GEIGER: %d\n",count_time_between);
```
ThomasWaldmann: \r\n ?

PR #162, opened 2020-01-18
Release V1.12.0

ThomasWaldmann:
doing final tests, will create a github release afterwards...


ThomasWaldmann:
@rexfue shall the release binary use the userdefines-example.h settings?

They are configured to send data to madavi and sensors.community.

rexfue:
I think that's OK. Madavi for quick tests and sensor.community for normal use (after registration there). The other defines also seem to be OK for me.

PR #171, opened 2020-01-19
mention release in read.me

rexfue:


ThomasWaldmann:
OK.

Later we can also mention the binary releases (esp. after we can do settings via web ui, so only one binary per hw device is needed).

rexfue:
Yes, that's the goal.

PR #178, opened 2020-02-03
WIP: add optional support for WS2812 RGB LED, fixes #106

### LINKED ISSUE #106: RFC: idea: connect a WS2812 LED
If the geiger is deployed outside in its grey plastik casing, there is little use for:

- the small onboard LED (pin 25 on all controllers)
- onboard OLED display
- maybe also the onboard piezo (except if one is very close with ones ears)

So, what remains is whatever it publishes on the internet.

For the geiger operator (and everybody else close-by) it would be nice to have a more direct display / feedback mechanism - this is how the WS2812 idea came up.

These LEDs are RGB programmable and exist as single LEDs (I got some 8mm ones) and also as LED strips (all with only 1 data input). Links (only to illustrate):

- https://www.sparkfun.com/products/12877
- https://www.sparkfun.com/products/12661

They need: Vcc (5V), GND, DATA_IN (usually works with high level from 3.3V output)
They provide: DATA_OUT for next LED, if any

So, for outside geigers, we could connect Pin25 to one WS2812B LED and let it shine in whatever colour or pulse pattern we like. So the geiger would have a nice shiny upper end. :-)

Mounting options:

- the (single) LED could stick through a whole in the upper end cap and get sealed by the glue that gets put on there anyway. The end cap would stay inside the grey tube and only the LED would stick out a little at the top.

- maybe even more pretty / noticable: let the white end cap stick out from the grey tube a bit and keep the single LED completely inside, so it will illuminate the end cap as a whole from the inside. the glue sealing would only happen on the side of the end cap in that case (not on the top). no drilling required.

The required software should be readily available via a library, these LEDs are quite popular - so our part is quite easy.

What do you think about it?

ThomasWaldmann:
specific indexes in the LED color time sequence are used for:

- dark separator indicates "start of sequence"
  (also used for LED init and re-init)

- color according to indications (3bits encoded via RGB color)
  (can be extended to show up to 32 indication bits)

- white separator indicates "start of radiation display"

- color according to radiation (repeated, displayed most of the time)
  dark (unusually low / 0)
  green (normal)
  yellow (more than normal)
  red (much more than normal)

**Still working on this, do not merge yet.**


multigeiger/multigeiger.ino
```diff
@@ -532,6 +537,9 @@ void loop()
 
     Dose_Rate = Count_Rate *GMC_factor_uSvph;                        // ... and dose rate
 
+    // indicate status on RGB LED (if any)
+    indicate(Dose_Rate, I_TEST);
```
ThomasWaldmann: just put it there for now.

in the end, i guess it should end up in a block that gets triggered every second.

ThomasWaldmann: The TEST indication is also just for now for testing and should be replaced by a indication computed from other geiger status values, e.g. whether HV generation has problems, WLAN/LoRa has problems, etc.

multigeiger/status_led.cpp
```diff
@@ -0,0 +1,105 @@
...
+        case 0:  // show a fixed dark separator after LED init
+            init_LED();  // (re-)init the LED
+            col = black;
+            break;
+        case 1:
+            r = indication & I_CONN_ERROR;
+            g = indication & I_TEST;
+            b = indication & I_HV_ERROR;
+            col = RgbColor(r ? 255 : 0, g ? 255 : 0, b ? 255 : 0);
+            break;
```
ThomasWaldmann: blocks like this could be also added for 2, 3, ... to show more indications.

multigeiger/status_led.cpp
```diff
@@ -0,0 +1,105 @@
...
+                col = red;
+            } else
+            if(radiation > 0.2) {
+                col = yellow;
+            } else
+            if(radiation > 0.01) {
+                col = green;
+            } else {
+                col = black;
+            }
```
ThomasWaldmann: Just hardcoded some values for now. Final thresholds to be discussed / defined.

multigeiger/status_led.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    set_LED(col);
+
+    if(++index == CSL)
+        index = 0;
```
ThomasWaldmann: First wanted to use modulo, but this should be usually faster.

multigeiger/status_led.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#else // no STATUS_LED
+
+void indicate(float radiation, unsigned int indication) {
+    // do nothing
+}
```
ThomasWaldmann: Due to this dummy we do not need lots of preprocessor boiler code at each place where we call indicate().

ThomasWaldmann: There could be also serial output here (if there is no LED / LED is switched off).

ThomasWaldmann:
rebased on master, fixed some led init issue.

on my hw it still shows wrong colors, maybe a hw issue.

t-pi:
Just had a quick view. This looks quite complex for a simple LED alarm (if that was the case) - and I admit I don't remember what we discussed a year ago 😄 

Proposition: 
- Simplify to basically calling `indicate_led` (or similar) once every `publish`
- Calculate the color directly from the dose rate, with
  - red starting from now existing alarm threshold
  - yellow starting from (alarm threshold - 1/factor), to reflect safety margin represented by alarm factor
- Use a more direct color calculation. Example from CO2 alarm see below. This shifts from green over yellow to red, and the handy 256 blocks can certainly be implemented more flexibly to reflect the config.

CO2 color example:
```
long get_CO2color(int CO2ppm) {
    switch (CO2ppm) {
      case 0 ... 700:
        // 0xGGRRBB
        return 0xFF0000;
      case 701 ... 956:
        return (0xFF0000+((CO2ppm-701)<<8));
      case 957 ... 1211:
        return ((0xFFFF00-((CO2ppm-957)<<16)));
      case 1212 ... 1500:
        return (0x00FF00);
      default:
        return (0x00FFFF);
    }
}
```

ThomasWaldmann:
The reason why it looks complex is that it is not only indicating one value (radiation), but also some more state infos from.

So I implemented a small state machine to signal more information serially.

Some color computation like you showed for co2 might be useful. I'll try it as soon as my led shows the colors I want.

PR #179, opened 2020-02-04
try both adresses of BME280

ecocurious:
There are BME280 boards with different addresses (0x76 and 0x77). Try to init the BME280 on both addresses.


rexfue:
This PR is from me. I was logged in as ecocurius.

PR #182, opened 2020-02-09
add a basic, short README in English

ThomasWaldmann:
this is about how a project README should look like, IMHO.

it should contain enough information for the reader to decide whether
this is the project they searched for.

but it should not be TL;DR (== too long; didn't read).

thus, it should NOT contain detailled installation and configuration
docs (that stuff belongs into the extensive docs the README links to).

this README has 2 uses / functions:
- being included as introduction into the long readthedocs.io based docs
- being shown on github (after review / community decision)

this avoids duplication and eases future changes/updates.

### DIFF
--- a/README-en.rst
+++ b/README-en.rst
@@ -1,0 +1,54 @@
+What is MultiGeiger?
+--------------------
+
+The MultiGeiger is a radioactivity measurement device.
+
+Main features
+~~~~~~~~~~~~~
+
+**multiple use cases supported**
+
+ * fixed location stations (data transfer via network)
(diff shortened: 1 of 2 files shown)


rexfue:
That is a good idea. 
Looks good for me, but i don't think Nicola will agree.

ThomasWaldmann:
I'll merge it anyway.

Considering that we have no english readme at all currently, this is an improvement.

If Nicola wants to improve it further, she can do so any time.

ThomasWaldmann:
https://multigeiger.readthedocs.io/en/latest/

PR #186, opened 2020-02-11
new log(level, format, ...) function, fixes #161

### LINKED ISSUE #161: refactor logging
currently, serial/usb logging is done usually via `Serial.printf()` or so.

i think we could replace this by an own `log(formatstring, ...)` function, that does:
- prepend "GEIGER: " to the format string
- append "\r\n" to the format string
- call Serial.printf(format_string, ...)

pros:
- less to type
- more consistent / less error prone (prefix and line endings)

con:
- it is only single-line that way (not big issue, hopefully)

Alternatively (or as a future enhancement), we could also go to a logging class / logger instance, that does:
- logging levels (INFO, WARNING, ERROR, CRITICAL) of log messages
- logging level of logger (== minimum level required to actually output a message)
- maybe different prefixes for header/formatting lines and real data lines

ThomasWaldmann:
- printf-style call
- only outputs for level >= log_level
- levels: DEBUG, INFO, WARNING, ERROR, CRITICAL, NOLOG
- internally prepends "GEIGER: " to differentiate our output from other
- adds correct line endings by using Serial.println()

not tested yet.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -70,6 +70,13 @@
 // STICK -> Heltec Wireless Stick (has LoRa on board)
 #define STICK 2
 
+// log levels
+#define DEBUG 0
+#define INFO 1
+#define WARNING 2
+#define ERROR 3
+#define CRITICAL 4
+#define NOLOG 999  // only to set log_level, so log() never creates output
 
 // Includes
 //====================================================================================================================================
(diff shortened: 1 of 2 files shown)


multigeiger/multigeiger.ino
```diff
@@ -888,7 +906,7 @@ String buildhttpHeaderandBodySBM(HTTPClient *head, unsigned int hvpulses, unsign
   body += ",{\"value_type\":\"sample_time_ms\",\"value\":\"" + String(timediff) + "\"}";
   body += "]}";
   if (DEBUG_SERVER_SEND == 1) {
-    Serial.println(body);
+    log(DEBUG, "body: %s", body.c_str());
```
ThomasWaldmann: not sure if `.c_str()` is needed here (and at similar places).

rexfue:
What about switching off logging completely?
I suggest to have a #define in userdefines.h to switch the debug level (and OFF).

ThomasWaldmann:
Well, when using the existing functionality in this PR, switching logging off is just raising log_level very high, so nothing gets through (I can add a level to do that).

I can add a #define for DEFAULT_LOG_LEVEL.

rexfue:
OK, I think that's a good idea

rexfue:
Yeah, looks good.

ThomasWaldmann:
TODO:
- practical test
- review log levels in `log()` calls

rexfue:
testet different log levels -> works as expected
level in log() calls seem to be OK
pls merge

PR #193, opened 2020-02-11
improve development docs

ThomasWaldmann:
some links, some new content, better formatting.

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -37,7 +37,8 @@ Some guidance for contributors:
   - have a nice, clear, typo-free commit comment
   - if you fixed an issue, refer to it in your commit comment
 
-- make a pull request on github
+- make a pull request on github and check on the PR page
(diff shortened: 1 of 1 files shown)


docs/source/development.rst
```diff
@@ -61,25 +68,32 @@ Documentation
...
   make html
 
-Then point a web browser at docs/build/html/index.html.
+Then point a web browser at ``docs/build/html/index.html``.
 
 The website is updated automatically by ReadTheDocs through GitHub web hooks on the
 main repository.
 
+.. _Sphinx: https://www.sphinx-doc.org/
+
```
ThomasWaldmann: @rexfue Do we need to add a link to https://docutils.sourceforge.io/rst.html or can it be found easily enough on the sphinx homepage?

rexfue: OK, I found it. Others should find it too - so no link necessary.

PR #196, opened 2020-02-12
use bump2version

ThomasWaldmann:
this tool can be used to automatically update the version numbers in misc. files (currently we have 3 places), so that this does not have to be done manually.

also it adds a corresponding git tag and, as configured now, signs the tag.

this avoids mistakes like forgetting some places, doing inconsistent updates or forgetting to set a tag.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -38,15 +38,6 @@
 //
 //
 
-// Release version and date
-// To fit in 16bit for lora version number we have some limits here:
-// VERSION_MAJOR = 0..15, VERSION_MINOR = 0..255, VERSION_PATCH = 0..15.
-#define VERSION_MAJOR 1
-#define VERSION_MINOR 12
-#define VERSION_PATCH 0
-// Date is in format "YYYY-MM-DD".
-#define VERSION_DATE "2020-01-18"
-
 // Fix Parameters
 // Values for Serial_Print_Mode to configure Serial (USB) output mode.  DON'T TOUCH!
 #define Serial_None 0            // No Serial output
@@ -80,6 +71,7 @@
 
 // Includes
 //====================================================================================================================================
+#include "version.h"
 #include "userdefines.h"
 #include <Arduino.h>
 #include <U8x8lib.h>
@@ -258,7 +250,6 @@ const char *Serial_Logging_Body = "%10d %15d %10f %9f %9d %8d %9d %9f %9f";
 const char *Serial_One_Minute_Log_Header = "%4s %10s %29s";
 const char *Serial_One_Minute_Log_Body = "%4d %10d %29d";
 const char *Serial_Logging_Name = "Simple Multi-Geiger";
-char revString[25];
 unsigned int lora_software_version;
 const char *dashes = "-------------------------------------------------------------------------------------------------";
 int Serial_Print_Mode = SERIAL_DEBUG;
@@ -419,13 +410,12 @@ void setup() {
   // just for fun
   log(INFO, "Let's go!");
 
-  // build revString
-  sprintf(revString, "V%d.%d.%d %s", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, VERSION_DATE);
-
   #if SEND2LORA
-  // build LoRa software version
-  lora_software_version = (VERSION_MAJOR << 12) + (VERSION_MINOR << 4) + VERSION_PATCH;
+  int major, minor, patch;
+  sscanf(VERSION_STR, "V%d.%d.%d", &major, &minor, &patch);
+  lora_software_version = (major << 12) + (minor << 4) + patch;
   #endif
+
   // Check, if we have a BME280 connected:
   haveBME280 = bme.begin(BME280_ADDRESS);
   if (haveBME280 == 0) {
@@ -460,7 +450,7 @@ void setup() {
 
   if (Serial_Print_Mode == Serial_Logging) {
     log(INFO, dashes);
-    log(INFO, "%s, Version %s", Serial_Logging_Name, revString);
+    log(INFO, "%s, Version %s", Serial_Logging_Name, VERSION_STR);
     log(INFO, dashes);
     log(INFO, Serial_Logging_Header,
         "GMC_counts", "Time_difference", "Count_Rate", "Dose_Rate", "HV Pulses", "Accu_GMC", "Accu_Time", "Accu_Rate", "Accu_Dose");
@@ -471,7 +461,7 @@ void setup() {
 
   if (Serial_Print_Mode == Serial_One_Minute_Log) {
     log(INFO, dashes);
-    log(INFO, "%s, Version %s", Serial_Logging_Name, revString);
+    log(INFO, "%s, Version %s", Serial_Logging_Name, VERSION_STR);
     log(INFO, dashes);
     log(INFO, Serial_One_Minute_Log_Header,
         "Time", "Count_Rate", "Counts");
@@ -482,7 +472,7 @@ void setup() {
 
   if (Serial_Print_Mode == Serial_Statistics_Log) {
     log(INFO, dashes);
-    log(INFO, "%s, Version %s", Serial_Logging_Name, revString);
+    log(INFO, "%s, Version %s", Serial_Logging_Name, VERSION_STR);
     log(INFO, dashes);
     log(INFO, "Time between two impacts");
     log(INFO, "[usec]");
@@ -762,13 +752,13 @@ void DisplayStartscreen(void) {
   u8x8.drawString(0, 2, "Geiger-");
   u8x8.drawString(0, 3, " Counter");
   u8x8.drawString(0, 4, "Version:");
-  sprintf(line, "%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
+  snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
   u8x8.drawString(0, 5, line);
   #else
   u8x8.setFont(u8x8_font_7x14_1x2_f);
   u8x8.println("Geiger-Counter");
   u8x8.println("==============");
-  sprintf(line, "V%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
+  snprintf(line, 15, "%s", VERSION_STR);  // 14 chars + \0 termination
   u8x8.setCursor(7 - strlen(line) / 2, 4);
   u8x8.print(line);
   u8x8.setCursor(1, 6);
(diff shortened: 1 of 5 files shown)


ThomasWaldmann:
TODO: refactor the code so it derives all required values (see TODO in the code) from the VERSION_STR.

rexfue:
Can we then use an extra version.h file with the version number? I want to use the version in my GUI, an if its an extra h-file I can easily include it.

rexfue:
OK, changes on VERSION_STR can be made, no problem with that.

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
 
   u8x8.clear();
```
ThomasWaldmann: the above line clears the display.

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
...
   #if CPU == STICK
-  // Display is only 4 lines by 8 characters; lines counting from 2 to 5
-  u8x8.setFont(u8x8_font_5x8_f);        // use really small font
-  for (int i = 2; i < 6; i++) {
-    u8x8.drawString(0, i, "        ");  // clear all 4 lines
-  }
```
ThomasWaldmann: this cleared the display again!? removed that.

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
...
+  // Display is only 4 lines by 8 characters; lines counting from 2 to 5.
+  snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
+  u8x8.setFont(u8x8_font_5x8_f);
   u8x8.drawString(0, 2, "Geiger-");
   u8x8.drawString(0, 3, " Counter");
   u8x8.drawString(0, 4, "Version:");
-  sprintf(line, "%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
   u8x8.drawString(0, 5, line);
   #else
+  // Display is only H lines by W characters; lines counting from A to B. -- TODO
```
ThomasWaldmann: @rexfue do you know or do we need to find out?

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
 
   u8x8.clear();
   #if CPU == STICK
-  // Display is only 4 lines by 8 characters; lines counting from 2 to 5
-  u8x8.setFont(u8x8_font_5x8_f);        // use really small font
-  for (int i = 2; i < 6; i++) {
-    u8x8.drawString(0, i, "        ");  // clear all 4 lines
-  }
+  // Display is only 4 lines by 8 characters; lines counting from 2 to 5.
```
ThomasWaldmann: the docs say X and Y both start from 0.

so, the docs are wrong? or do we want to leave 2 empty lines at the top?

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
...
+  snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
+  u8x8.setFont(u8x8_font_5x8_f);
   u8x8.drawString(0, 2, "Geiger-");
   u8x8.drawString(0, 3, " Counter");
   u8x8.drawString(0, 4, "Version:");
-  sprintf(line, "%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
   u8x8.drawString(0, 5, line);
   #else
+  // Display is only H lines by W characters; lines counting from A to B. -- TODO
+  snprintf(line, 15, "%s", VERSION_STR);  // 14 chars + \0 termination
```
ThomasWaldmann: the stuff below is 14 chars wide, so I assumed this is the max. for the version, too.

multigeiger/multigeiger.ino
```diff
@@ -754,25 +751,21 @@ void DisplayStartscreen(void) {
...
-  sprintf(line, "V%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
-  u8x8.setCursor(7 - strlen(line) / 2, 4);
-  u8x8.print(line);
-  u8x8.setCursor(1, 6);
-  u8x8.print("Info:boehri.de");
+  u8x8.drawString(0, 2, "Geiger-Counter");
+  u8x8.drawString(0, 3, "==============");
+  u8x8.drawString(0, 4, line);
+  u8x8.drawString(0, 6, "Info:boehri.de");
   #endif
```
ThomasWaldmann: be consistent with the code for the small display: use drawString for the bigger display also.

multigeiger/multigeiger.ino
```diff
@@ -419,13 +417,12 @@ void setup() {
   // just for fun
   log(INFO, "Let's go!");
 
-  // build revString
-  sprintf(revString, "V%d.%d.%d %s", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, VERSION_DATE);
```
ThomasWaldmann: will now only be the full VERSION_STR (including "-dev" for development versions).

ThomasWaldmann:
TODO:
- practically test initial OLED display output on WIFI and STICK devices.
- check version on lora data (should be ok, but better check)
- check serial output (should be ok also)

rexfue:
checked OLEDs - see comments above
checked LoRa - OK
checked serial - OK

On the small display we should omit the "V" and replace it with a space. If one day we have versions above 9, there will be no more space.

ThomasWaldmann:
I don't see any "comments above". Did you submit them?

Having a major release usually means breaking / major changes, these do not happen too often, so reaching major release 10 will likely take 10-20 years for us, if ever.

So, for the simplicity of the code, we'll solve that when/if we reach 10 and keep the V for now. :-)

rexfue:
Sorry - forget to submit.

Agreed on the "V" - by the I may no longer be here ...

multigeiger/multigeiger.ino
```diff
@@ -754,25 +744,21 @@ void DisplayStartscreen(void) {
...
+  snprintf(line, 15, "%s", VERSION_STR);  // 14 chars + \0 termination
   u8x8.setFont(u8x8_font_7x14_1x2_f);
-  u8x8.println("Geiger-Counter");
-  u8x8.println("==============");
-  sprintf(line, "V%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
-  u8x8.setCursor(7 - strlen(line) / 2, 4);
-  u8x8.print(line);
-  u8x8.setCursor(1, 6);
-  u8x8.print("Info:boehri.de");
+  u8x8.drawString(0, 2, "Geiger-Counter");
```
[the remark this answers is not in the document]
ThomasWaldmann: ok, i'll check/fix.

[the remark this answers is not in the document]
ThomasWaldmann:
```
  u8x8.drawString(0, 2, "Geiger-Counter");
  u8x8.drawString(0, 3, "==============");
  u8x8.drawString(0, 4, line);
  u8x8.drawString(0, 6, "Info:boehri.de");
```

0,2,4,6 can't be correct - I guess we want to have the underlining directly under the "Geiger-Counter" string.

1,2,4,6 maybe?

multigeiger/multigeiger.ino
ThomasWaldmann: Isn't the library expected to be generic for all supported displays?

So, if it doesn't clear, file a bug against the library?

ThomasWaldmann: And does it still happen with the latest lib release?

multigeiger/multigeiger.ino
ThomasWaldmann: And the result for H and W is ... ?

multigeiger/multigeiger.ino
ThomasWaldmann: Again: file a bug against the library? It's docs say 0/0.

multigeiger/multigeiger.ino
ThomasWaldmann: OK, H=4, W=16.

multigeiger/multigeiger.ino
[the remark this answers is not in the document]
ThomasWaldmann: Ehrm, above you said H is 4.

So does line 0 and 1 overlap or why do you say 0, 2, 4, 6 (normally that would mean H==7)?

multigeiger/multigeiger.ino
```diff
@@ -384,8 +375,17 @@ char *buildSSID() {
...
   u8x8.begin();
+  // On the WIFI CPU u8x8.begin() cleared the display.
+  #if CPU == STICK
+  // But not on the STICK CPU, so we do it here.
+  // Display is only 4 lines by 8 characters; lines counting from 2 to 5
+  u8x8.setFont(u8x8_font_5x8_f);        // use really small font
+  for (int i = 2; i < 6; i++) {
+    u8x8.drawString(0, i, "        ");  // clear all 4 lines
+  }
+  #endif
```
ThomasWaldmann: if u8x8.begin() clears the WIFI display, but not the STICK display, doing both here should be structurally better, right?

multigeiger/multigeiger.ino
```diff
@@ -419,13 +419,12 @@ void setup() {
...
   log(INFO, "Let's go!");
 
-  // build revString
-  sprintf(revString, "V%d.%d.%d %s", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, VERSION_DATE);
-
   #if SEND2LORA
-  // build LoRa software version
-  lora_software_version = (VERSION_MAJOR << 12) + (VERSION_MINOR << 4) + VERSION_PATCH;
+  int major, minor, patch;
+  sscanf(VERSION_STR, "V%d.%d.%d", &major, &minor, &patch);
```
ThomasWaldmann: why is major, minor coloured differently here than patch?

multigeiger/multigeiger.ino
```diff
@@ -751,28 +750,23 @@ int gen_charge_pulses(int max_charge_pulses) {
...
+  snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
+  u8x8.setFont(u8x8_font_5x8_f);
   u8x8.drawString(0, 2, "Geiger-");
   u8x8.drawString(0, 3, " Counter");
   u8x8.drawString(0, 4, "Version:");
-  sprintf(line, "%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
   u8x8.drawString(0, 5, line);
   #else
+  // Display is only 4 lines by 16 characters; lines counting 0,2,4,6.
+  snprintf(line, 17, "%s", VERSION_STR);  // 16 chars + \0 termination
```
ThomasWaldmann: using up to 16 chars here now.

multigeiger/multigeiger.ino
```diff
@@ -751,28 +750,23 @@ int gen_charge_pulses(int max_charge_pulses) {
...
-  u8x8.println("==============");
-  sprintf(line, "V%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
-  u8x8.setCursor(7 - strlen(line) / 2, 4);
-  u8x8.print(line);
-  u8x8.setCursor(1, 6);
-  u8x8.print("Info:boehri.de");
+  u8x8.drawString(0, 0, "Geiger-Counter");
+  u8x8.drawString(0, 2, "==============");
+  u8x8.drawString(0, 4, line);
+  u8x8.drawString(0, 6, "Info:boehri.de");
```
ThomasWaldmann: this is only 14 wide. we could add some chars:
- blanks left and right from Geiger-Counter
- 2x `=` more
- a blank between `Info:` and `boehri.de`

ThomasWaldmann:
ok, read the u8x8 docs.

0,2,4,6 is correct for the double-height font (because the raster is always 8x8 and does not adjust to font size).

still unclear why it starts from 2 and clear does not work for the smaller display.

ThomasWaldmann:
I reverted all display-related changes, they didn't really fit in here.

ThomasWaldmann:
I'm merging this. As I removed all display-related changes anyway, this is now much simpler.

PR #200, opened 2020-02-15
modularize code

ThomasWaldmann:


multigeiger/tube.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void setup_tube() {
+  pinMode(PIN_TEST_OUTPUT, OUTPUT);
+  pinMode(PIN_HV_FET_OUTPUT, OUTPUT);
+  pinMode(PIN_HV_CAP_FULL_INPUT, INPUT);
+  pinMode(PIN_GMC_COUNT_INPUT, INPUT);
+
+  digitalWrite(PIN_TEST_OUTPUT, LOW);
+  digitalWrite(PIN_HV_FET_OUTPUT, LOW);
```
ThomasWaldmann: this line was missing in the original code.

multigeiger/tube.cpp
```diff
@@ -0,0 +1,116 @@
...
+  do {
+    digitalWrite(PIN_HV_FET_OUTPUT, HIGH);              // turn the HV FET on
+    delayMicroseconds(1500);                            // 5000 usec gives 1,3 times more charge, 500 usec gives 1/20 th of charge
+    digitalWrite(PIN_HV_FET_OUTPUT, LOW);               // turn the HV FET off
+    delayMicroseconds(1000);
+    charge_pulses++;
+  } while ((charge_pulses < max_charge_pulses) && !isr_GMC_cap_full); // either a timeout or a capacitor full interrupt stops this loop
+  time2hvpulse = millis();                              // we just pulsed, so restart timer
+  if ((charge_pulses == max_charge_pulses) && !isr_GMC_cap_full)
+    return -1; // TODO log(CRITICAL, "HV charging failed!");               // pulsed a lot, but still the capacitor is not at desired voltage
```
ThomasWaldmann: TODO

multigeiger/tube.cpp
```diff
@@ -0,0 +1,116 @@
...
+// How many HV capacitor charge pulses to generate before giving up.
+// (MAX_CHARGE_PULSES * pulse_duration) should be less than the interval in
+// the main loop (currently 1000ms) where it unconditionally recharges.
+#define MAX_CHARGE_PULSES 333
+
+// How many HV capacitor charge pulses to generate for first charge,
+// for the call in setup(). Not timing critical and capacitor is empty.
+#define MAX_CHARGE_PULSES_INITIAL 3333
+
+int gen_charge_pulses(bool setup) {
```
ThomasWaldmann: to be able to have the pulse count constants here, i changed the gen_charge_pulses api so it only gets a parameter whether it is called from setup() or not. From setup() we want to do many pulses, but not from elsewhere.

multigeiger/tube.cpp
```diff
@@ -0,0 +1,112 @@
...
+  hvpulse_timestamp = millis();
+  if ((charge_pulses == max_charge_pulses) && !isr_GMC_cap_full)
+    return -1; // TODO log(CRITICAL, "HV charging failed!");               // pulsed a lot, but still the capacitor is not at desired voltage
+  return charge_pulses;
+}
+
+void setup_tube() {
+  pinMode(PIN_TEST_OUTPUT, OUTPUT);
+  pinMode(PIN_HV_FET_OUTPUT, OUTPUT);
+  pinMode(PIN_HV_CAP_FULL_INPUT, INPUT);
```
ThomasWaldmann: this line was missing in the original code.

multigeiger/display.cpp
```diff
@@ -0,0 +1,151 @@
...
+  sprintf(format, "%%%dd", digits);
+  sprintf(erg, format, n);
+  return erg;
+}
+
+void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool connected) {
+  if (!use_display && !displayIsClear) {
+    u8x8.clear();
+    clearDisplayLine(4);
+    clearDisplayLine(5);
```
ThomasWaldmann: @rexfue do you know why the lower 2 lines are needed?

multigeiger/loraWan.cpp
```diff
@@ -1,7 +1,26 @@
-#ifndef LORAWAN_H
-#define LORAWAN_H
```
ThomasWaldmann: such stuff is usually found in the `.h` file (thus the LORAWAN_H name) - why was it here?

multigeiger/switches.cpp
```diff
@@ -0,0 +1,41 @@
...
+  sw[0] = !digitalRead(PIN_SWI_0);
+  sw[1] = !digitalRead(PIN_SWI_1);
+  sw[2] = !digitalRead(PIN_SWI_2);
+  sw[3] = !digitalRead(PIN_SWI_3);
+
+  // log state of switch
+  log(DEBUG, "SW0: %d  SW1: %d  SW2: %d  SW3: %d", sw[0], sw[1], sw[2], sw[3]);
+
+  return sw;
+}
```
ThomasWaldmann: would be a bit cleaner to return an integer (bit field), with each bit representing a switch.
for now i kept the code similar to as it was, maybe doing that later...

multigeiger/thp_sensor.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <Adafruit_BME280.h>
+
+#include "log.h"
+#include "thp_sensor.h"
+
+int haveBME280 = 0;
+
+float bme_temperature = 0.0;
+float bme_humidity = 0.0;
+float bme_pressure = 0.0;
```
ThomasWaldmann: we should remove the "bme(280)" later.

the sensor name does not matter for the temperature, pressure, humidity values.

multigeiger/transmission.cpp
```diff
@@ -0,0 +1,182 @@
...
+  http.addHeader("X-PIN", "19");
+  tube_type = tube_type.substring(10);
+  String valuetype = (addname ? tube_type + "_" : "");
+  valuetype += "counts_per_minute";
+  String body = "{\"software_version\":\"" + http_software_version + "\",\"sensordatavalues\":[";
+  body += "{\"value_type\":\"" + valuetype + "\",\"value\":\"" + cpm + "\"}";
+  body += ",{\"value_type\":\"hv_pulses\",\"value\":\"" + String(hv_pulses) + "\"}";
+  body += ",{\"value_type\":\"counts\",\"value\":\"" + String(gm_counts) + "\"}";
+  body += ",{\"value_type\":\"sample_time_ms\",\"value\":\"" + String(timediff) + "\"}";
+  body += "]}";
```
ThomasWaldmann: also needs improvement, but kept it mostly as is for now.

multigeiger/transmission.cpp
```diff
@@ -0,0 +1,182 @@
...
+  prepare_http(&http, host);
+  http.addHeader("X-PIN", "11");
+  String temp = (addname ? "BME280_temperature" : "temperature");
+  String humi = (addname ? "BME280_humidity" : "humidity");
+  String press = (addname ? "BME280_pressure" : "pressure");
+  String body = "{\"software_version\":\"" + http_software_version + "\",\"sensordatavalues\":[\
+    {\"value_type\":\"" + temp + "\",\"value\":\"" + String(temperature, 2) + "\"},\
+    {\"value_type\":\"" + humi + "\",\"value\":\"" + String(humidity, 2) + "\"},\
+    {\"value_type\":\"" + press + "\",\"value\":\"" + String(pressure, 2) + "\"}\
+  ]}";
```
ThomasWaldmann: also needs improvement, but kept it mostly as is for now.

multigeiger/transmission.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int hv_pulses, unsigned int gm_counts, unsigned int cpm,
+                   int have_thp, float temperature, float humidity, float pressure) {
+
+  #if SEND2DUMMY
+  displayStatusLine("Toilet");
+  log(INFO, "SENDING TO TOILET ...");
+  send_http_geiger(TOILET, tube_type, dt, hv_pulses, gm_counts, cpm);
+  if (have_thp)
+    send_http_thp(TOILET, temperature, humidity, pressure);
```
ThomasWaldmann: not using globals any more.

ThomasWaldmann:
Tested:
- WIFI board display output
- log output at INFO level

Not tested:
- sending to servers

multigeiger/display.cpp
rexfue: I guess, it was because lines 4 and 5 where not cleared by u8x8.clear() at small display.

multigeiger/tube.cpp
rexfue: Which on of this tree ?

ThomasWaldmann: Sorry, I don't understand the question.

multigeiger/display.cpp
ThomasWaldmann: but that would be an upstream bug in the library.

for such stuff a bug should be filed against the library code.

multigeiger/loraWan.cpp
```diff
@@ -96,59 +117,54 @@ const lmic_pinmap lmic_pins = {
...
-void onEvent (ev_t ev) {
-    Serial.print(os_getTime());
-    Serial.print(": ");
-    switch(ev) {
+void onEvent(ev_t ev) {
+    switch (ev) {
         case EV_SCAN_TIMEOUT:
-            Serial.println(F("EV_SCAN_TIMEOUT"));
             txStatus = TX_STATUS_ENDING_ERROR;
+            event_log(F("EV_SCAN_TIMEOUT"));
```
rexfue: You can (should) leave out the F(...)
on ESP32 ist is not necessary, ESP32 always puts const (literal) strings into flash and handles it transparently

multigeiger/userdefines-example.h
```diff
@@ -1,67 +1,76 @@
...
 //
 // To use them please copy this file to userdefines.h (do not add that file to version control)
 // and then edit userdefines.h as appropriate for your use case / hardware.
-//====================================================================================================================================
 
-// If you use platformio, CPU is defined in platformio.ini and
-// you don't need to change the definition of CPU.
-// Arduino IDE users *must* define CPU here and uncomment one and only one of following CPU-Defines.
+
+// CPU values (board types, DO NOT CHANGE):
```
rexfue: Can we put all these "DO NOT CHANGE" defines in an extra file? 
I think it's dangerous that a user CAN change it.

rexfue:
I'm testing lora and the sending to the servers.

multigeiger/loraWan.cpp
ThomasWaldmann: yeah, can clean that up later. for this PR will just keep it as it was before already.

rexfue: I didnt compile here with the F(
I removed them and it worked

multigeiger/userdefines-example.h
ThomasWaldmann: well, if you edit the source (which is always the case except when using a precompiled binary), you can always shoot yourself in the foot.

I can put the constants into yet another header file and #include that there (as I did for some other stuff in userdefines.h).

But then, one has to tell the user their options (constant names), so that is a duplication that has to be kept in sync with the header file.

rexfue: OK agreed.
Then we could cluster this DONT CHANGE together and make  an attention line like ***************************** DONT CHANGE HERE ****************** or similar

multigeiger/loraWan.cpp
ThomasWaldmann: Oh? It compiled for me and it also compiled on travis-ci.

So is your setup somehow different?

ThomasWaldmann: (hmm, ignore the "for me" case, I did not compile for lora hw)

but on travis, we compile for wifi and for lora hw.

rexfue: Here's the error message:

multigeiger/loraWan.cpp: In function 'void onEvent(ev_t)':
multigeiger/loraWan.cpp:133:43: error: cannot convert 'const __FlashStringHelper*' to 'const char*' for argument '1' to 'void event_log(const char*)'
             event_log(F("EV_SCAN_TIMEOUT"));

multigeiger/transmission.cpp
```diff
@@ -0,0 +1,202 @@
...
+ "sensordatavalues": [
+  {"value_type": "%scounts_per_minute", "value": "%d"},
+  {"value_type": "hv_pulses", "value": "%d"},
+  {"value_type": "counts", "value": "%d"},
+  {"value_type": "sample_time_ms", "value": "%d"}
+ ]
+}
+)=====";
+  snprintf(body, 1000, json_format,
+           http_software_version,
```
rexfue: Either add here .c_str()  (-> http_software_version.c_str() ) or declare http_software_version as char* .
The %s cannot be used with String !
A few lines down it comes again.

multigeiger/loraWan.cpp
```diff
@@ -96,59 +117,54 @@ const lmic_pinmap lmic_pins = {
...
 };
 
-static volatile transmitionStatus_t txStatus;
+static volatile transmissionStatus_t txStatus;
 static uint8_t * __rxPort;
 static uint8_t * __rxBuffer;
 static uint8_t * __rxSz;
 
+void event_log(const char *event){
+  log(DEBUG, "%s: %s", os_getTime(), event);
```
rexfue: os_getTime doesn't return a char*, it returns ostime_t. so it  should be  
log(DEBUG, "%d: %s", os_getTime(), event);
Question: why do you read and show the time here?

rexfue:
Really good job !!

Found 2 problems - see comments.
LoRa works and sending to servers too.

multigeiger/userdefines-example.h
```diff
@@ -1,67 +1,76 @@
...
+#define WIFI 0  // Heltec Wifi Kit 32
+#define LORA 1  // Heltec Wifi Lora 32 (V2)
```
rexfue: This must be deleted - this board is no longer supported.
We only have WifiKit32 and Wireless Stick

multigeiger/display.cpp
ThomasWaldmann: https://github.com/ecocurious2/MultiGeiger/issues/208

multigeiger/switches.cpp
ThomasWaldmann: https://github.com/ecocurious2/MultiGeiger/issues/209

multigeiger/tube.cpp
ThomasWaldmann: rephrase? oder auf deutsch?

rexfue: Versteh nur Bahnhof.
Du schreibst oben: this line was missing in ...
Und ich frage: which one of this three  (gut, ich hab das e von one vergessen)
Ich weiß nicht, welche Zeile gefehlt hat.
Aber das ist ja eigentlich auch völlig unwichtig.

multigeiger/loraWan.cpp
ThomasWaldmann: OK, I configured for STICK and repective hw, now I also see that. Fixing...

multigeiger/transmission.cpp
ThomasWaldmann: right, fixing...

multigeiger/loraWan.cpp
ThomasWaldmann: the code already was like that, I only changed the direct call to Serial.print* to log(...).

ThomasWaldmann: fixing the format...

multigeiger/userdefines-example.h
ThomasWaldmann: should be done after this PR (not in scope).

multigeiger/loraWan.cpp
```diff
@@ -296,11 +309,10 @@ transmitionStatus_t lorawan_send(uint8_t txPort, uint8_t * txBuffer, uint8_t txS
                 return txStatus;
           }
```
ThomasWaldmann: sketch/loraWan.cpp: In function 'transmissionStatus_t lorawan_send(uint8_t, uint8_t*, uint8_t, bool, uint8_t*, uint8_t*, uint8_t*)':
loraWan.cpp:299:18: error: enumeration value 'TX_STATUS_ENDING_ERROR' not handled in switch [-Werror=switch]
           switch ( txStatus ) {

why am I suddenly seeing this? i did not change this code block.

how should it be handled?

rexfue: I don't know - here it is only a warning !

multigeiger/loraWan.cpp: In function 'transmissionStatus_t lorawan_send(uint8_t, uint8_t*, uint8_t, bool, uint8_t*, uint8_t*, uint8_t*)':
multigeiger/loraWan.cpp:299:18: warning: enumeration value 'TX_STATUS_ENDING_ERROR' not handled in switch [-Wswitch]
           switch ( txStatus ) {
                  ^
multigeiger/loraWan.cpp:299:18: warning: enumeration value 'TX_STATUS_TIMEOUT' not handled in switch [-Wswitch]

You could change the -Werror=switch to -Wswitch

ThomasWaldmann: fixed compiler warning without changing behaviour. should be checked nevertheless...

multigeiger/tube.cpp
ThomasWaldmann: die direkt ueber meinem kommentar.

multigeiger/loraWan.cpp
ThomasWaldmann: fixed.

multigeiger/transmission.cpp
ThomasWaldmann: fixed.

multigeiger/loraWan.cpp
ThomasWaldmann: fixed

multigeiger/loraWan.cpp
ThomasWaldmann: Having "compiler warning: all" in Arduino IDE. Seems to transform some warnings to errors...

multigeiger/userdefines-example.h
ThomasWaldmann: #211

multigeiger/userdefines-example.h
ThomasWaldmann: I just placed the stuff in a way so that you first see your options and directly after that, you configure what you have. Then next options, next configuration. Repeat.

Guess every way has its pros and cons.

ThomasWaldmann:
@rexfue guess it's ready now?

can you do another test with the current code?

the only todo i still have is maybe to do another rebase -i master with this and collapse some changesets (but no code changes).

rexfue:
OK, i'll do the tests - maybe tomorrow or saturday.

ThomasWaldmann:
rebase -i and collapsing some changesets is done.

ready to merge after testing.

rexfue:
OK, testet. Lora OK, send to server OK, compiling in platform and on Arduino OK
All tests made with wireless stick
After merge i will change server data a little - especially for MADAVI

ThomasWaldmann:
\o/

PR #215, opened 2020-02-21
remove bme/BME from variable names, fixes #210

### LINKED ISSUE #210: remove bme(280) from variable / function names
we should remove the "bme(280)".

the sensor name does not matter for the temperature, pressure, humidity values.

so only hw-specific bme280 code should keep it.

ThomasWaldmann:


rexfue:
If you really think, that that must be, then merge.

ThomasWaldmann:
yeah, I do think so.

we might be some day using a BME999 or some totally different sensor, who knows.

PR #216, opened 2020-02-21
use a bitfield for the dipswitches, fixes #209

### LINKED ISSUE #209: use bitfield for switches
https://github.com/ecocurious2/MultiGeiger/pull/200#discussion_r380385383

ThomasWaldmann:


multigeiger/multigeiger.ino
```diff
@@ -115,12 +115,12 @@ void setup() {
 void loop() {
   unsigned long time_difference;
   unsigned int HV_pulse_count;
-  char *sw;
+  Switches switches;
```
ThomasWaldmann: sw was a bit short anyway, when i searched for that i regularly found password also...

multigeiger/switches.h
```diff
@@ -4,9 +4,14 @@
 #define _SWITCHES_H_
 
 // What are the switches good for?
-enum {SPEAKER_ON, DISPLAY_ON, LED_ON, UNUSED};
+typedef struct switches {
+  unsigned int speaker_on: 1;  // SW0
+  unsigned int display_on: 1;  // SW1
+  unsigned int led_on: 1;  // SW2
+  unsigned int unused: 1;  // SW3
+} Switches;
```
ThomasWaldmann: sometimes even C has nice things.

rexfue: Yea, look good.
But is it better ini terms of code size?

ThomasWaldmann: I didn't check, but I would not expect more than a few bytes difference, if any.

ThomasWaldmann: master:

856326 bytes (25%) of program storage.
Global variables use 44912 bytes (13%) of dynamic memory.

this PR:

856338 bytes (25%) of program storage space.
Global variables use 44912 bytes (13%) of dynamic memory.

So, 12 bytes more program code.

rexfue: OK, that's negligible.

PR #218, opened 2020-02-22
adapt_server_data

rexfue:
send to MADAVI in one single request both geiger and thp

add sensor name if sending to MADAVI
define magic numbers for XPIN


multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
   const char *json_format = R"=====(
 {
  "software_version": "%s",
  "sensordatavalues": [
-  {"value_type": "%stemperature", "value": "%.2f"},
-  {"value_type": "%shumidity", "value": "%.2f"},
-  {"value_type": "%spressure", "value": "%.2f"}
+  {"value_type": "temperature", "value": "%.2f"},
+  {"value_type": "humidity", "value": "%.2f"},
+  {"value_type": "pressure", "value": "%.2f"}
```
ThomasWaldmann: that's better, yeah! :)

multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
+           pressure);
   send_http(&http, body);
 }
 
+void send_http_geiger_and_thp(const char *host, String tube_type, unsigned int timediff, 
+  unsigned int hv_pulses, unsigned int gm_counts, unsigned int cpm,
+  float temperature, float humidity, float pressure) {
+  char body[1000];
+  HTTPClient http;
+  prepare_http(&http, host, 0);
```
ThomasWaldmann: pin 0?

multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
+{
+ "software_version": "%s",
+ "sensordatavalues": [
+  {"value_type": "%s_counts_per_minute", "value": "%d"},
+  {"value_type": "%s_hv_pulses", "value": "%d"},
+  {"value_type": "%s_counts", "value": "%d"},
+  {"value_type": "%s_sample_time_ms", "value": "%d"},
+  {"value_type": "BME280_temperature", "value": "%.2f"},
+  {"value_type": "BME280_humidity", "value": "%.2f"},
+  {"value_type": "BME280_pressure", "value": "%.2f"}
```
ThomasWaldmann: remove bme280 here also?

multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
+  prepare_http(&http, host, 0);
+  tube_type = tube_type.substring(10);
+  const char *json_format = R"=====(
+{
+ "software_version": "%s",
+ "sensordatavalues": [
+  {"value_type": "%s_counts_per_minute", "value": "%d"},
+  {"value_type": "%s_hv_pulses", "value": "%d"},
+  {"value_type": "%s_counts", "value": "%d"},
+  {"value_type": "%s_sample_time_ms", "value": "%d"},
```
ThomasWaldmann: see comment above, sensor name should have extra key/value pair.

multigeiger/transmission.cpp
```diff
@@ -165,27 +196,30 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
   #endif
 
   #if SEND2MADAVI
   log(INFO, "Sending to Madavi ...");
   displayStatusLine("Madavi");
-  send_http_geiger(MADAVI, tube_type, dt, hv_pulses, gm_counts, cpm);
-  if (have_thp)
-    send_http_thp(MADAVI, temperature, humidity, pressure);
+  if (have_thp) {
+    // send both infos in one request to MADAVI to keep the server load lower
```
ThomasWaldmann: btw, doing 2 requests is maybe not as bad as one thinks.

if we keep the connection, it is just a 2nd request flowing over existing tcp (and in future tls) connection.

i tried that when experimenting with https, the 2nd+ request is much faster than the first.

multigeiger/transmission.h
```diff
@@ -5,6 +5,13 @@
 #ifndef _TRANSMISSION_H_
 #define _TRANSMISSION_H_
 
+// Sensor-PINS.
+// They are called PIN, because in the first days of Feinstaub sensor they were
+// really the CPU-Pins. Now they are 'virtual' pins to distinguish different sensors.
+// Since we send to sensor.community, we have to use their numbers.
+#define XPIN_RADIATION 19
+#define XPIN_BME280 11
```
ThomasWaldmann: good to have the explanation here!

if this is specific to sensor.community **only**, the pin stuff should be removed from prepare_http so that that can be used as common code for any json request, without us sending a header that is neither needed nor understood by other servers.

multigeiger/transmission.cpp
```diff
@@ -46,11 +46,12 @@ void setup_transmission(const char *version, char *ssid) {
...
 }
 
-void prepare_http(HTTPClient *http, const char *host) {
+void prepare_http(HTTPClient *http, const char *host, int xpin) {
   http->begin(host);
   http->addHeader("Content-Type", "application/json; charset=UTF-8");
   http->addHeader("Connection", "close");
   http->addHeader("X-Sensor", chipID);
+  http->addHeader("X-PIN", String(xpin));
 }
```
ThomasWaldmann: please don't do X-Pin here, this is intended to be generic code to prepare a json http request, no matter which server we are talking to.

multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
 
-void send_http_thp(const char *host, float temperature, float humidity, float pressure) {
-  bool addname = false;
-  String prefix = (addname ? "BME280_" : "");
+void send_http_thp(const char *host, float temperature, float humidity, float pressure, int xpin) {
   char body[1000];
   HTTPClient http;
-  prepare_http(&http, host);
-  http.addHeader("X-PIN", "11");
```
ThomasWaldmann: this should stay here. or we even need specific / different code for sensors.community and madavi.

multigeiger/transmission.cpp
```diff
@@ -70,59 +71,89 @@ void send_http(HTTPClient *http, String body) {
...
 )=====";
   snprintf(body, 1000, json_format,
            http_software_version.c_str(),
            prefix.c_str(), cpm,
-           hv_pulses,
-           gm_counts,
-           timediff);
+           prefix.c_str(), hv_pulses,
+           prefix.c_str(), gm_counts,
+           prefix.c_str(), timediff);
```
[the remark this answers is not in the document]
rexfue: The names ARE important here for MADAVi. This server checks the name  to decide, which sensor it is and where to put the data in ist database.
And MADAVI builds the names from the value_type. We cannot change this.

multigeiger/transmission.cpp
rexfue: MADAVI doesn't interpret the X-PIN header so the number can be any

multigeiger/transmission.cpp
rexfue: NO

multigeiger/transmission.cpp
rexfue: See comment above - no.

multigeiger/transmission.cpp
rexfue: OK, may be. But Rajko told me, i shall use one request!

multigeiger/transmission.cpp
rexfue: OK, understood. Will change that.

multigeiger/transmission.cpp
rexfue: Yes, I guess ist would be best to use different code.

multigeiger/transmission.cpp
ThomasWaldmann: but then the header rather should not be there.

rexfue: Yeah, would be better.

multigeiger/transmission.cpp
ThomasWaldmann: if the same applies to X-Sensor, please move that also.

rexfue: Hmmm, this is the only point in the whole transmission, where the sensor ID is sent, so I think we need this for every server.
Btw. how can i get rid of the unwanted files here? I have adjusted my .gitignore so that they are not committed next time.

ThomasWaldmann: guess you could use `git rebase -i master` and then "e"dit the commit.

then, I guess it is `git rm --cached ...`.

multigeiger/transmission.cpp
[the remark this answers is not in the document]
ThomasWaldmann: ok, too bad.

multigeiger/transmission.cpp
rexfue: ok, he told me I SCHOULD use one request, but two are also OK. So I will stay on two calls, code is simpler then. And most of our geigers don‘t have a BME anyway.

rexfue:
Don't know why loran.cpp is in this PR - git is magic.

ThomasWaldmann:
You have some of my commits in your PR.

Maybe do that:
```
git checkout master  # change to your local master
git pull git@github.com:ecocurious2/MultiGeiger.git master  # update your local master
git checkout adapt_server_data  # change to this PR branch again
git rebase -i master  # reapply your changes on top of current master
```
Within the last operation, you will also see what changeset you have in that branch.

You can choose to remove or reorder changesets then, read the comment in the editor you get shown.

rexfue:
Thanks, I'll try that

rexfue:
Did that, but:

rxf@MacBig-2718 MultiGeiger % git push
To https://github.com/rexfue/MultiGeiger.git
 ! [rejected]        adapt_server_data -> adapt_server_data (non-fast-forward)
error: failed to push some refs to 'https://github.com/rexfue/MultiGeiger.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

ThomasWaldmann:
If you are sure you want to push what you have, use `git push -f`.

It does not work without -f because it is not appending new stuff at the end, but doing potentially destructive changes (which is expected after using git rebase -i).

ThomasWaldmann:
https://github.com/ecocurious2/MultiGeiger/pull/218/files still has unwanted files in it.

multigeiger/transmission.h
```diff
@@ -5,6 +5,14 @@
 #ifndef _TRANSMISSION_H_
 #define _TRANSMISSION_H_
 
+// Sensor-PINS.
+// They are called PIN, because in the first days of Feinstaub sensor they were
+// really the CPU-Pins. Now they are 'virtual' pins to distinguish different sensors.
+// Since we send to sensor.community, we have to use their numbers.
+#define XPIN_NO_XPIN 0
```
ThomasWaldmann: so 0 does not exist as a PIN?

multigeiger/transmission.cpp
```diff
@@ -91,35 +91,84 @@ void send_http_geiger(const char *host, String tube_type, unsigned int timediff,
...
+)=====";
+  snprintf(body, 1000, json_format,
+           http_software_version.c_str(),
+           temperature,
+           humidity,
+           pressure);
+  send_http(&http, body);
+}
+
+// two extra functiions for MADAVI, because MADAVI needs the sensorname in value_type to recignise the sensors
```
ThomasWaldmann: typo recognize

multigeiger/transmission.h
rexfue: Yes because it was a hardware pin number and they always start from 1

rexfue:
I agree with your proposal to create a new release after this (and your PRs) are merged.
The 'settings via GUI' stuff can be done after that for the next release.

ThomasWaldmann:
if you like, you could use `git rebase -i master` to cleanup your commits:
- move a fix commit directly below the commit it is fixing
- use f(ixup) or s(quash)

rexfue:
OK - will do it next time.
Can i merge?

PR #219, opened 2020-02-22
multigeiger cleanups

ThomasWaldmann:
lots of cleanups, guess best reviewed by looking at the individual commits.


rexfue:
I would leave speaker_ticks etc as variable. If we change to set them using the web server, we need them as variable.

ThomasWaldmann:
speaker_ticks as a variable is not needed as of now.

if we get a settings module or an extended webconf module in future, it can be easily re-added there (together with a `extern ...` declaration in the respective header file).

ThomasWaldmann:
merge?

rexfue:
Yeah

PR #221, opened 2020-02-22
use less globals

ThomasWaldmann:


ThomasWaldmann:
rebased onto current master, will merge this "as is" and do more work on globals after next release.

rexfue:
OK, looks good.

PR #226, opened 2020-03-02
Use webpage to to set configuration, fixes #140

### LINKED ISSUE #140: Settings via web GUI
The settings in userdefines.h should be adjustable via web interface. At least a part of it.
Started to work on that.

rexfue:
lorawan.cpp:
 - only compile, if CPU==STICK, i.e if we have LoRa Hardware
 - fetch LoRa credentials from config variables

multigeiger.ino:
 - use config variables instead of userdefines

transmission.cpp:
 - replace SEND2LORA by CPU==STICK
 - use config variables to decide where to send
 - send to LoRa only, if LoRa credentials are known

webconf.cpp
 - add all config variable to web page
 - do restart after saving config

userdefines-example.h
 - remove now unnecessary #defines

### DIFF
--- a/multigeiger/loraWan.cpp
+++ b/multigeiger/loraWan.cpp
@@ -19,6 +19,8 @@
 // DIO2 (10) = GPIO32 (12)
 
 #include "userdefines.h"
+#include "webconf.h"
+#include "utils.h";
 
 // Send a valid LoRaWAN packet using frequency and encryption settings matching
 // those of the The Things Network.
@@ -282,39 +284,6 @@ transmissionStatus_t lorawan_send(uint8_t txPort, uint8_t *txBuffer, uint8_t txS
       }
     }
   }
-}
-
-//convert hexstring to len bytes of data
-//returns 0 on success, -1 on error
-//data is a buffer of at least len bytes
-//hexstring is upper or lower case hexadecimal, NOT prepended with "0x"
-int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse) {
-  const char *pos = hexstring;
-  char *endptr;
-  size_t count = 0;
-  if ((hexstring[0] == '\0') || (strlen(hexstring) % 2)) {
-    //hexstring contains no data
-    //or hexstring has an odd length
-    return -1;
-  }
-  for (count = 0; count < len; count++) {
-    char buf[5] = {'0', 'x', pos[0], pos[1], 0};
-    data[count] = strtol(buf, &endptr, 0);
-    pos += 2 * sizeof(char);
-    if (endptr[0] != '\0') {
-      //non-hexadecimal character encountered
-      return -1;
-    }
-  }
-  if (reverse) {
-    char temp;
-    for (int i = 0; i < len / 2; i++) {
-      temp = data[i];
-      data[i] = data[len - i - 1];
-      data[len - i - 1] = temp;
-    }
-  }
-  return 0;
 }
 
 #endif
--- a/multigeiger/loraWan.cpp
+++ b/multigeiger/loraWan.cpp
@@ -1,3 +1,6 @@
+// to have CPU defined:
+#include "userdefines.h"
+
 // Compile this only if we have a LoRa capable hardware
 #if CPU==STICK
 
@@ -17,10 +20,6 @@
 // DIO0 (8) = GPIO26 (15)
 // DIO1 (9) = GPIO33 (13)
 // DIO2 (10) = GPIO32 (12)
-
-#include "userdefines.h"
-#include "webconf.h"
-#include "utils.h";
 
 // Send a valid LoRaWAN packet using frequency and encryption settings matching
 // those of the The Things Network.
@@ -47,12 +46,9 @@
 #include "hal/heltecv2.h"
 #include <SPI.h>
 #include "webconf.h"
-
+#include "utils.h"
 #include "loraWan.h"
 #include "log.h"
-
-// Prototype
-int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse);
 
 // For normal use, we require that you edit the sketch to replace FILLMEIN
 // with values assigned by the TTN console. However, for regression tests,
@@ -70,15 +66,17 @@ int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool 
 // All these 3 LoRa parameters (DEVEUI, APPEUI and APPKEY) may be copied literally from the TTN console window.
 // The necessary reversal on DEVEUI and APPEUI is done by hex2data.
 void os_getArtEui(u1_t *buf) {
-  (void) hex2data(buf, (const char *) appeui, 8, true);
+  (void) hex2data(buf, (const char *) appeui, 8);
+  reverseByteArray(buf, 8);
 }
 
 void os_getDevEui(u1_t *buf) {
-  (void) hex2data(buf, (const char *) deveui, 8, true);
+  (void) hex2data(buf, (const char *) deveui, 8);
+  reverseByteArray(buf, 8);
 }
 
 void os_getDevKey(u1_t *buf) {
-  (void) hex2data(buf, (const char *) appkey, 16, false);
+  (void) hex2data(buf, (const char *) appkey, 16);
 }
 
 static osjob_t sendjob;
--- a/multigeiger/loraWan.cpp
+++ b/multigeiger/loraWan.cpp
@@ -1,3 +1,6 @@
+// Compile this only if we have a LoRa capable hardware
+#if CPU==STICK
+
 // loraWan hardware related code
 // code based on free-to-use / do-anything-what-you-want-with-it example code
 // by Thomas Telkamp, Matthijs Kooijman and Terry Moore, MCCI.
@@ -17,13 +20,6 @@
 
 #include "userdefines.h"
 
-// Check if a CPU (board) with LoRa is selected. If not, deactivate SEND2LORA.
-#if !((CPU==LORA) || (CPU==STICK))
-#undef SEND2LORA
-#define SEND2LORA 0
-#endif
-
-#if SEND2LORA
 // Send a valid LoRaWAN packet using frequency and encryption settings matching
 // those of the The Things Network.
 //
@@ -48,9 +44,13 @@
 #include <hal/hal.h>
 #include "hal/heltecv2.h"
 #include <SPI.h>
+#include "webconf.h"
 
 #include "loraWan.h"
 #include "log.h"
+
+// Prototype
+int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse);
 
 // For normal use, we require that you edit the sketch to replace FILLMEIN
 // with values assigned by the TTN console. However, for regression tests,
@@ -64,29 +64,19 @@
 # define FILLMEIN (#dont edit this, edit the lines that use FILLMEIN)
 #endif
 
-// This EUI must be in little-endian format, so least-significant-byte
-// first. When copying an EUI from ttnctl output, this means to reverse
-// the bytes. For TTN issued EUIs the last bytes should be 0xD5, 0xB3, 0x70.
-// static const u1_t PROGMEM APPEUI[8]={ 0x45, 0xB6, 0x01, 0xD0, 0x7E, 0xD5, 0xB3, 0x70};
-static const u1_t PROGMEM APPEUI[8] = { 0x98, 0x7A, 0x01, 0xD0, 0x7E, 0xD5, 0xB3, 0x70 };
+
+// All these 3 LoRa parameters (DEVEUI, APPEUI and APPKEY) may be copied literally from the TTN console window.
+// The necessary reversal on DEVEUI and APPEUI is done by hex2data.
 void os_getArtEui(u1_t *buf) {
-  memcpy_P(buf, APPEUI, 8);
+  (void) hex2data(buf, (const char *) appeui, 8, true);
 }
 
-// This should also be in little endian format, see above.
-//static const u1_t PROGMEM DEVEUI[8]={ 0x00, 0x02, 0x25, 0xB0, 0x03, 0xCC, 0x10, 0x03 };
-static const u1_t PROGMEM DEVEUI[8] = { 0xE8, 0x7C, 0x19, 0xC3, 0x00, 0xC0, 0xD0, 0x00 };
 void os_getDevEui(u1_t *buf) {
-  memcpy_P(buf, DEVEUI, 8);
+  (void) hex2data(buf, (const char *) deveui, 8, true);
 }
 
-// This key should be in big endian format (or, since it is not really a
-// number but a block of memory, endianness does not really apply). In
-// practice, a key taken from ttnctl can be copied as-is.
-//static const u1_t PROGMEM APPKEY[16] = { 0x16, 0x77, 0xD0, 0x1A, 0xEB, 0x57, 0xE5, 0x28, 0x84, 0x23, 0x19, 0x12, 0x70, 0xEC, 0xEF, 0xF0 };
-static const u1_t PROGMEM APPKEY[16] = { 0x6F, 0xA9, 0x57, 0x78, 0xF0, 0x9E, 0xC1, 0xC5, 0xB6, 0xAF, 0x88, 0x74, 0x35, 0x58, 0xF1, 0x87 };
 void os_getDevKey(u1_t *buf) {
-  memcpy_P(buf, APPKEY, 16);
+  (void) hex2data(buf, (const char *) appkey, 16, false);
 }
 
 static osjob_t sendjob;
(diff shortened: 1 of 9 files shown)


multigeiger/webconf.cpp
```diff
@@ -5,8 +5,47 @@
...
+// WiFi hardware: send to wifi servers and not to LoRa
+char sendToLora[2] = "0";
+char sendToCommunity[2] = "1";
+char sendToMadavi[2] = "1";
+#endif
+
+IotWebConfSeparator sep0 = IotWebConfSeparator("Tick settings");
+IotWebConfParameter startSoundParam = IotWebConfParameter("Start sound (1=ON)", "startSound", startSound, 2, "number", "0/1", NULL, "min='0' max='1' step='1'");
+IotWebConfParameter speakerTickParam = IotWebConfParameter("Speaker tick (1=ON)", "speakerTick", speakerTick, 2, "number", "0/1", NULL, "min='0' max='1' step='1'");
+IotWebConfParameter ledTickParam = IotWebConfParameter("LED ticker (1=ON)", "ledTick", ledTick, 2, "number", "0/1", NULL, "min='0' max='1' step='1'");
```
ThomasWaldmann: LED tick
(consistency)

multigeiger/webconf.cpp
```diff
@@ -90,3 +149,11 @@ void setup_webconf() {
   });
 }
 
+// Check value of settings parameters
+boolean parameterTrue(char *parameter) {
+  return (strcmp(parameter, "1") == 0);
+}
```
ThomasWaldmann: don't call this from all over the place, but rather use bool variables.

multigeiger/webconf.h
```diff
@@ -6,9 +6,40 @@
...
+extern IotWebConfSeparator sep1;
+extern IotWebConfParameter sendToCommunityParam;
+extern IotWebConfParameter sendToMadaviParam;
+#if CPU==STICK
+extern IotWebConfSeparator sep3;
+extern IotWebConfParameter sendToLoraParam;
+extern IotWebConfParameter appeuiParam;
+extern IotWebConfParameter deveuiParam;
+extern IotWebConfParameter appkeyParam;
+#endif
```
ThomasWaldmann: why are these IotWebConfParameter here?

multigeiger/webconf.cpp
```diff
@@ -80,6 +120,25 @@ void setup_webconf() {
   iotWebConf.getWifiSsidParameter()->label = "WiFi client SSID";
   iotWebConf.getWifiPasswordParameter()->label = "WiFi client password";
 
+  // add the setting parameter
```
ThomasWaldmann: maybe rather omit such comments, that's quite obvious from the code already.

multigeiger/webconf.cpp
```diff
@@ -80,6 +120,25 @@ void setup_webconf() {
...
+  iotWebConf.addParameter(&sep0);
+  //iotWebConf.addParameter(&startSoundParam1);
```
ThomasWaldmann: remove?

multigeiger/loraWan.cpp
```diff
@@ -293,4 +283,38 @@ transmissionStatus_t lorawan_send(uint8_t txPort, uint8_t *txBuffer, uint8_t txS
     }
   }
 }
+
+//convert hexstring to len bytes of data
+//returns 0 on success, -1 on error
+//data is a buffer of at least len bytes
+//hexstring is upper or lower case hexadecimal, NOT prepended with "0x"
+int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse) {
```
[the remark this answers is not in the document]
ThomasWaldmann: hmm, stdlib stuff likely can't support reverse. can this be solved differently or separately?

multigeiger/multigeiger.ino
```diff
@@ -66,7 +66,7 @@ void setup() {
   setup_webconf();
   setup_transmission(VERSION_STR, ssid);
 
-  if (PLAY_SOUND)
+  if (parameterTrue(startSound))
```
ThomasWaldmann: calling parameterTrue() is slightly ugly and always causing overhead. use a bool variable.

multigeiger/webconf.cpp
```diff
@@ -5,8 +5,47 @@
...
-
-#define CONFIG_VERSION "012"  // for IoTWebConfig
+#include "userdefines.h"
+
+// All these flags are ON (=1) at startup
+// They all can be changed at any time using the web interface
+char speakerTick[2] = "1";
+char startSound[2] = "1";
+char ledTick[2] = "1";
+char showDisplay[2] = "1";
```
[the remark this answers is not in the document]
ThomasWaldmann: BTW, this is where / why you still need SPEAKER_TICK value if you do not want to just hardcode "1" (true) here.

multigeiger/webconf.cpp
```diff
@@ -67,7 +106,8 @@ void handleRoot(void) {  // Handle web requests to "/" path.
 }
 
 void configSaved(void) {
-  log(INFO, "Config saved");
+  log(INFO, "Config saved .. restarting");
+  ESP.restart();
```
ThomasWaldmann: maybe we could detect actual config change somehow? maybe supported by iotWebConf already?

and only restart if changed?

multigeiger/loraWan.cpp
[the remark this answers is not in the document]
rexfue: Separately sure, differently may be. Do you have a good solution?

multigeiger/webconf.cpp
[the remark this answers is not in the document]
rexfue: To use a 'load_config()_ is a good idea. 
But I still don't see the advantage of a #define SPEAKER_TICK. You have to set the config variables to the the value of the #define at startup anyway.

multigeiger/webconf.h
rexfue: OK, i see, this is not needed

multigeiger/webconf.cpp
rexfue: Better more than too little comment ...

multigeiger/webconf.cpp
rexfue: OK

multigeiger/multigeiger.ino
rexfue: I think that results in an unnecessary duplication. You have to copy the bool variables to the config variables and back again.

multigeiger/webconf.cpp
rexfue: Didn't find a way to read the stored configs back from iotwebconf. May be you find one.

multigeiger/loraWan.cpp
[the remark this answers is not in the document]
ThomasWaldmann: hmm, crap, scanf %x is hex string -> integer... (not byte sequence).

i miss python. :-)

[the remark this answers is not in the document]
ThomasWaldmann: https://gist.github.com/xsleonard/7341172 that looks ... interesting.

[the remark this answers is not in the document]
rexfue: interesting - yeah. But he uses malloc. I don't like malloc. You have to free memory after that. And i don't know, why he puts a '\0' at the end - it's a byte array, no string - isn't it?

Concerning python: rewrite the project using micro python ! ;-)

multigeiger/multigeiger.ino
ThomasWaldmann: usually when dealing with user input one needs some validation step (and rejection if invalid, letting the user try again).
and as the validation has to deal with types and valid values/ranges anyway, it usually can transform the string to the desired type also.

multigeiger/webconf.cpp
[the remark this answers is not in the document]
ThomasWaldmann: the value is to have them in one place in a header file (and also to keep existing docs that describe this stuff).

multigeiger/webconf.cpp
ThomasWaldmann: no, it is not better if it does not add information - you have to read all the stuff, it produces longer source code.

sometimes it also has typos or sounds strange...

multigeiger/webconf.cpp
ThomasWaldmann: The `init` method does that via `loadConfig`.

I also found that one can set an own FormValidator.

multigeiger/multigeiger.ino
ThomasWaldmann: most codebases i have seen yet do a one-time read, validate, convert step when reading their text-based config (usually a text file in some format) and after that only use the appropriate types (bool, int, string, ...) but not the config text any more.

in our case here, the iotWebConf variables have to be considered our "text config" and the additional set of appropriately typed variables would contain the stuff we use from the loop.

if we immediately validate at the form input step and reject any form that has invalid data, we can immediately get the user to edit and resubmit the form, until we get input that validates.

rexfue: OK, sounds good. Would you please try that?

multigeiger/webconf.cpp
rexfue: Do you have another version of iotwebconf? There is no loadConfig. There is a configload, but that is private.

ThomasWaldmann: https://github.com/prampec/IotWebConf/blob/master/src/IotWebConf.cpp#L116

yeah, configLoad is private, but init is public.

multigeiger/utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+    data[count] = strtol(buf, &endptr, 0);
+    pos += 2 * sizeof(char);
+    if (endptr[0] != '\0') {
+      //non-hexadecimal character encountered
+      return -1;
+    }
+  }
+  if (reverse) {
+    reverseByteArray(data, len);
+  }
```
ThomasWaldmann: removing this would be cleaner. just call both functions from the higher layer.

multigeiger/utils.h
```diff
@@ -0,0 +1,11 @@
...
+
+#ifndef UTILS_H
+#define UTILS_H
+
+// Prototypes
+int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse);
+void reverseByteArray(unsigned char *data, int len);
+
+
+#endif
```
ThomasWaldmann: add a linefeed at the end so diff does not complain.

multigeiger/webconf.cpp
```diff
@@ -7,41 +7,40 @@
 #include <HTTPClient.h>
 #include "userdefines.h"
 
-// All these flags are ON (=1) at startup
-// They all can be changed at any time using the web interface
-char speakerTick[2] = "1";
-char startSound[2] = "1";
-char ledTick[2] = "1";
-char showDisplay[2] = "1";
+bool speakerTick = (SPEAKER_TICK == 1);
```
ThomasWaldmann: have SPEAKER_TICK as a boolean so this expression is not needed?

multigeiger/webconf.cpp
```diff
@@ -105,9 +104,38 @@ void handleRoot(void) {  // Handle web requests to "/" path.
...
+
+void loadConfigVariables(void) {
+  // check if WiFi SSID has changed. If so, restart cpu. Otherwise, the program will not use the new SSID
+  if ((strcmp(lastWiFiSSID, "") != 0) && (strcmp(lastWiFiSSID, iotWebConf.getWifiSsidParameter()->valueBuffer) != 0)) {
+    log(INFO, "Doing restart...");
+    ESP.restart();
+  }
+  strcpy(lastWiFiSSID, iotWebConf.getWifiSsidParameter()->valueBuffer);
+
+  speakerTick = strcmp(speakerTick_c, "1") == 0;
```
ThomasWaldmann: ok for now, i'ld add something a bit better later.

rexfue: OK

multigeiger/webconf.cpp
rexfue: ?? SPEAKER_TICK is a #define, so it is an integer , isn't it?

multigeiger/utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+  return 0;
+}
+
+// reverse a bytearray of length len
```
ThomasWaldmann: "only works for even lengths"

multigeiger/utils.h
```diff
@@ -0,0 +1,11 @@
+// Div. utilities
```
ThomasWaldmann: see above.

multigeiger/webconf.cpp
ThomasWaldmann: a #define is a preprocessor text replacement, so it can be a lot of things.

multigeiger/webconf.cpp
```diff
@@ -90,3 +185,6 @@ void setup_webconf() {
   });
 }
 
+
+
+
```
ThomasWaldmann: remove?

multigeiger/webconf.h
```diff
@@ -6,9 +6,24 @@
...
+extern char appeui[];
+extern char deveui[];
+extern char appkey[];
+#endif
+
 extern char ssid[];
 extern IotWebConf iotWebConf;
 
 void setup_webconf(void);
+boolean parameterTrue(char *parameter);
```
ThomasWaldmann: remove?

rexfue: OK

multigeiger/webconf.cpp
rexfue: OK

multigeiger/utils.cpp
rexfue: comment added

multigeiger/utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "utils.h"
+
+// ** convert hexstring to len bytes of data
+//returns 0 on success, -1 on error
+//data is a buffer of at least len bytes
+//hexstring is upper or lower case hexadecimal, NOT prepended with "0x"
+//convert hexstring to len bytes of data
+//returns 0 on success, -1 on error
+//data is a buffer of at least len bytes
+//hexstring is upper or lower case hexadecimal, NOT prepended with "0x"
```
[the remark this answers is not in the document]
rexfue: The style checker didn't complain :-(

multigeiger/utils.cpp
```diff
@@ -0,0 +1,49 @@
+// Div utilities
```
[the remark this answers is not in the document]
rexfue: Changed to misc

multigeiger/transmission.cpp
```diff
@@ -221,32 +216,34 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
-  #endif
 
-  #if SEND2LORA
-  log(INFO, "Sending to TTN ...");
-  displayStatusLine("TTN");
-  send_ttn_geiger(tube_nbr, dt, gm_counts);
-  if (have_thp)
-    send_ttn_thp(temperature, humidity, pressure);
+  #if CPU==STICK
+  if(sendToLora && (strcmp(appeui,"") != 0)) { // send only, if we have LoRa credentials
```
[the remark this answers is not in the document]
rexfue: No, this style checker doesn't work well, i guess.

multigeiger/loraWan.cpp
```diff
@@ -48,10 +46,14 @@
...
 #include "hal/heltecv2.h"
 #include <SPI.h>
+#include "webconf.h"
 
 #include "loraWan.h"
 #include "log.h"
 
+// Prototype
+int hex2data(unsigned char *data, const char *hexstring, unsigned int len, bool reverse);
+
```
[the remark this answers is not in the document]
rexfue: OK

multigeiger/loraWan.cpp
```diff
@@ -16,14 +19,9 @@
 // DIO2 (10) = GPIO32 (12)
 
 #include "userdefines.h"
+#include "webconf.h"
+#include "utils.h";
```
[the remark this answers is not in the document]
rexfue: Uhps

rexfue:
Sorry, commented already 6 days ago but forgot to click on comment. PR is out

multigeiger/transmission.cpp
```diff
@@ -221,32 +216,34 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
-  #if SEND2SENSORCOMMUNITY
-  log(INFO, "Sending to sensor.community ...");
-  displayStatusLine("sensor.community");
-  send_http_geiger(SENSORCOMMUNITY, dt, hv_pulses, gm_counts, cpm, XPIN_RADIATION);
-  if (have_thp) {
-    send_http_thp(SENSORCOMMUNITY, temperature, humidity, pressure, XPIN_BME280);
+  if(sendToCommunity) {
+    log(INFO, "Sending to sensor.community ...");
+    displayStatusLine("sensor.community");
+    send_http_geiger(SENSORCOMMUNITY, dt, hv_pulses, gm_counts, cpm, XPIN_RADIATION)  ;
```
ThomasWaldmann: remove the blanks before the semicolon, please.

multigeiger/utils.cpp
```diff
@@ -0,0 +1,46 @@
...
+// returns 0 on success, -1 on error
+// data is a buffer of at least len bytes
+// hexstring is upper or lower case hexadecimal, NOT prepended with "0x"
+int hex2data(unsigned char *data, const char *hexstring, unsigned int len) {
+  const char *pos = hexstring;
+  char *endptr;
+  size_t count = 0;
+  if ((hexstring[0] == '\0') || (strlen(hexstring) % 2)) {
+    //hexstring contains no data
+    //or hexstring has an odd length
```
ThomasWaldmann: // + 1 blank

ThomasWaldmann: (can you search and replace that globally?)

rexfue:
Fixed. Can you please make the 1.13 release after merging. Thanks.

rexfue:
Merged. You can make the 1.13 release.

PR #228, opened 2020-03-25
webconf cleanups

ThomasWaldmann:


multigeiger/webconf.cpp
```diff
@@ -28,22 +29,41 @@ char deveui[17] = "";
...
+    *value = true;
+    return true;
+  return false;  // invalid
+}
+
+void format_bool(bool *value, char *text) {
+  text[0] = *value ? '1' : '0';
+  text[1] = '\0';
+}
+
```
ThomasWaldmann: later, when we have web form input validation, we can also use different strings than "0" and "1" for the booleans, e.g. "false" and "true".

multigeiger/webconf.cpp
```diff
@@ -28,22 +29,41 @@ char deveui[17] = "";
...
+IotWebConfSeparator sep1 = IotWebConfSeparator("Transmission settings");
+IotWebConfParameter sendToCommunityParam = BOOL_PARAM("Send to sensors.community", "send2Community", sendToCommunity_c);
+IotWebConfParameter sendToMadaviParam = BOOL_PARAM("Send to madavi.de", "send2Madavi", sendToMadavi_c);
+
 #if CPU==STICK
-IotWebConfSeparator sep3 = IotWebConfSeparator("LoRa settings");
-IotWebConfParameter sendToLoraParam = IotWebConfParameter("Send to LoRa (=>TTN) (1=ON)", "send2lora", sendToLora_c, 2, "number", "0/1", NULL, "min='0' max='1' step='1'");
+IotWebConfSeparator sep2 = IotWebConfSeparator("LoRa settings");
+IotWebConfParameter sendToLoraParam = BOOL_PARAM("Send to LoRa (=>TTN)", "send2lora", sendToLora_c);
```
ThomasWaldmann: this one would fit better into the "transmission settings" section. of course only show if we have lora hw.

multigeiger/webconf.cpp
```diff
@@ -28,22 +29,40 @@ char deveui[17] = "";
...
+    return true;
+  if (!strcmp(text, "1"))
+    *value = true;
+    return true;
+  return false;  // invalid
+}
+
+void format_bool(bool *value, char *text) {
+  strcpy(text, *value ? "1" : "0");
+}
```
ThomasWaldmann: later, when we have web form input validation, we can also use different strings than "0" and "1" for the booleans, e.g. "false" and "true".

multigeiger/webconf.cpp
```diff
@@ -28,22 +29,40 @@ char deveui[17] = "";
...
 
+bool parse_bool(char *text, bool *value) {
+  if (!strcmp(text, "0"))
+    *value = false;
+    return true;
+  if (!strcmp(text, "1"))
+    *value = true;
+    return true;
+  return false;  // invalid
+}
```
rexfue: This won't work. This is NOT python. You need curly braces like:
  if (!strcmp(text, "0"))
     {
         *value = false;
         return true;
   }

multigeiger/webconf.cpp
rexfue: I guess 1/0 is better, because it's easier to type in on a small phone keyboard - it is no to error prone.

multigeiger/webconf.cpp
ThomasWaldmann: oops :-)

rexfue: OK now.

ThomasWaldmann:
rebased on master, squashed some fixups, merging now...

PR #231, opened 2020-03-30
check if LoRa chip or not by hardware

rexfue:
test, if pin 26 (LoRa DIO) is low. If yes, then it is LoRa hardware.
Change in  every file the "#if CPU == STICK" to "if (haveLoRa ) {}"

### DIFF
--- a/multigeiger/chkhardware.cpp
+++ b/multigeiger/chkhardware.cpp
@@ -1,0 +1,17 @@
+// Check if we have LoRa hardware
+
+// Test pin 26 (that is LORA_DIO0) to be low.
+#include <Arduino.h>
+
+#define HWTESTPIN 26
+
+// Test pin is LoRa_DIO0. If LoRa chip is present, this pin is low, otherwise
+// it is an open pin with pullup
+bool init_hwtest(void) {
+  pinMode(HWTESTPIN, INPUT_PULLUP);
+  delay(200);
+  if (!digitalRead(HWTESTPIN)) {      // low => LoRa chip detected
+    return true;
+  }
+  return false;
+}
--- a/multigeiger/chkhardware.h
+++ b/multigeiger/chkhardware.h
@@ -1,0 +1,6 @@
+#ifndef CHKHARDWARE_H
+#define CHKHARDWARE_H
+
+bool init_hwtest(void);
+
+#endif
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -9,81 +9,74 @@
 
 #include "display.h"
 
-#if CPU == STICK
 #define PIN_DISPLAY_ON 25
-#endif
 
 #define PIN_OLED_RST 16
 #define PIN_OLED_SCL 15
 #define PIN_OLED_SDA 4
 
-#if CPU != STICK
 U8X8_SSD1306_128X64_NONAME_HW_I2C u8x8(PIN_OLED_RST, PIN_OLED_SCL, PIN_OLED_SDA);
-#else
-U8X8_SSD1306_64X32_NONAME_HW_I2C u8x8(PIN_OLED_RST, PIN_OLED_SCL, PIN_OLED_SDA);
-#endif
+U8X8_SSD1306_64X32_NONAME_HW_I2C u8x8_lora(PIN_OLED_RST, PIN_OLED_SCL, PIN_OLED_SDA);
+U8X8 *pu8x8;
 
 bool displayIsClear;
+static bool isLoraBoard;
 
 void display_start_screen(void) {
   char line[20];
 
-  u8x8.clear();
-  #if CPU == STICK
-  // Display is only 4 lines by 8 characters; lines counting from 2 to 5
-  u8x8.setFont(u8x8_font_5x8_f);        // use really small font
-  for (int i = 2; i < 6; i++) {
-    u8x8.drawString(0, i, "        ");  // clear all 4 lines
+  pu8x8->clear();
+  if (isLoraBoard) {
+    // Display is only 4 lines by 8 characters; lines counting from 2 to 5
+    pu8x8->setFont(u8x8_font_5x8_f);        // use really small font
+    for (int i = 2; i < 6; i++) {
+      pu8x8->drawString(0, i, "        ");  // clear all 4 lines
+    }
   }
-  #endif
 
-  #if CPU == STICK
-  u8x8.drawString(0, 2, "Geiger-");
-  u8x8.drawString(0, 3, " Counter");
-  u8x8.drawString(0, 4, "Version:");
-  snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
-  u8x8.drawString(0, 5, line);
-  #else
-  u8x8.setFont(u8x8_font_7x14_1x2_f);
-  u8x8.drawString(0, 0, "Geiger-Counter");
-  u8x8.drawString(0, 2, "==============");
-  snprintf(line, 15, "%s", VERSION_STR);  // 14 chars + \0 termination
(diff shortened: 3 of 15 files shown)


rexfue:
A first try. Query the LoRa DIO0 pin. It works, tested with two different Heltec modules.

multigeiger/chkhardware.h
```diff
@@ -0,0 +1,6 @@
+#ifndef CHKHARDWARE_H
+#define CHKHARDWARE_H
+
+bool init_hwtest(void);
+
+#endif
```
ThomasWaldmann: add a linefeed to make diff happy.

multigeiger/display.cpp
```diff
@@ -9,81 +9,77 @@
...
-void setup_display(void) {
-  u8x8.begin();
-
-  #if CPU == STICK
-  pinMode(PIN_DISPLAY_ON, OUTPUT);
-  digitalWrite(PIN_DISPLAY_ON, HIGH);
-  #endif
-
+void setup_display(bool lora) {
+  haveLoRa = lora;
```
ThomasWaldmann: as far as this module is concerned, it is not really about having lora or not, it is about which kind of display we have.
which is also implied if we know which `board` we have.

multigeiger/switches.cpp
```diff
@@ -6,19 +6,18 @@
...
-#define PIN_SWI_3 36
-#endif
-
-void setup_switches() {
+static unsigned int PIN_SWI_0 = 39;
+static unsigned int PIN_SWI_1 = 38;
+static unsigned int PIN_SWI_2 = 37;
+static unsigned int PIN_SWI_3 = 36;
+
+void setup_switches(bool lora) {
```
ThomasWaldmann: similar here, it is rather the board hardware (aka "CPU") that influences this than lora.

multigeiger/switches.h
```diff
@@ -11,7 +11,8 @@ typedef struct switches {
   unsigned int unused: 1;  // SW3
 } Switches;
 
-void setup_switches(void);
+
+void setup_switches(bool);
```
ThomasWaldmann: bool lora (or whatever).

multigeiger/transmission.cpp
```diff
@@ -22,23 +19,24 @@
...
   http_software_version = String(version);
 
-  #if CPU==STICK
-  int major, minor, patch;
-  sscanf(version, "V%d.%d.%d", &major, &minor, &patch);
-  lora_software_version = (major << 12) + (minor << 4) + patch;
-  setup_lorawan();
-  #endif
+  if (haveLoRa) {
+    haveLoRa = true;
```
ThomasWaldmann: hmm?

multigeiger/transmission.cpp
```diff
@@ -236,15 +232,15 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
 
-  #if CPU==STICK
-  if(sendToLora && (strcmp(appeui, "") != 0)) {    // send only, if we have LoRa credentials
-    log(INFO, "Sending to TTN ...");
-    displayStatusLine("TTN");
-    send_ttn_geiger(tube_nbr, dt, gm_counts);
-    if (have_thp)
-      send_ttn_thp(temperature, humidity, pressure);
+  if (haveLoRa) {
+    if(sendToLora && (strcmp(appeui, "") != 0)) {    // send only, if we have LoRa credentials
```
ThomasWaldmann: just add another term to the second "if"?

multigeiger/webconf.h
```diff
@@ -14,15 +14,13 @@ extern bool sendToCommunity;
...
 extern char appeui[];
 extern char deveui[];
 extern char appkey[];
-#endif
 
 extern char ssid[];
 extern IotWebConf iotWebConf;
 
-void setup_webconf(void);
+void setup_webconf(bool);
```
ThomasWaldmann: bool lora (or whatever).

multigeiger/chkhardware.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+// Test pin is LoRa_DIO0. If LoRa chip is present, this pin is low, otherwise
+// it is an open pin with pullup
+bool init_hwtest(void) {
+  pinMode(HWTESTPIN, INPUT_PULLUP);       
+  delay(200);
+  if(!digitalRead(HWTESTPIN)) {       // low => LoRa chip detected
+    haveLoRa = true;
+  }  
+  return haveLoRa;
```
ThomasWaldmann:
```
return !digitalRead(HWTESTPIN);
```

also: either check this only once or kill that static variable, guess both together doesn't make sense.

multigeiger/chkhardware.cpp
```diff
@@ -0,0 +1,19 @@
+// Check if we have LoRa hardware
+
+// Test pin 26 (that is LORA_DIO0) to be low.
+#include <Arduino.h>
+
+#define HWTESTPIN 26
+
+static bool haveLoRa = false;
```
ThomasWaldmann: maybe this should be rather `board(_type)` (not "CPU" though as it was in userdefines.h, but with the same semantics as we had there: one specific integer identifying a specific board).

multigeiger/chkhardware.cpp
ThomasWaldmann: but a `int board` variable would be even better.

rexfue: I prefer a bool for easier tests and if/else and we only have 2 boards ( if there are sometimes more boards, we have to change the routine anyway). 
Let's call it 'bool loraBoard'. I want to have the 'lora' there , because we test the presence of the LoRa chip.

multigeiger/chkhardware.cpp
rexfue: see comment above

multigeiger/webconf.h
rexfue: That's not necessary say the C rules.

multigeiger/transmission.cpp
rexfue: OK

multigeiger/transmission.cpp
rexfue: typo - changed

multigeiger/chkhardware.h
rexfue: OK

multigeiger/webconf.h
ThomasWaldmann: yes, but we do it at other places that way.

multigeiger/chkhardware.cpp
ThomasWaldmann: I still think an enum is the more appropriate approach than a bool.

This is not about True/False, but about having a specific board or not (which could, even if we only have 2 now, be > 2 at any time soon [and was > 2 until recently also]).

ThomasWaldmann:
platformio env name fixed, but some (seen before) issues remain, see also there:

https://travis-ci.org/github/ecocurious2/MultiGeiger/jobs/667011487

ThomasWaldmann:
It's broken because RST_LoRa and DIOx constants are undefined for the Heltec Wifi Kit 32.

rexfue:
You don't have to compile WiFi32 - only compile Wireless Stick (as it is in paltformio.ini!). Wifi32 will be done in the code.

ThomasWaldmann:
fixed platformio build, rebased on master, squashed some commits.

ThomasWaldmann:
@rexfue did you flash the binary made for a "wifi stick" onto a "wifi kit 32" device and it worked?

rexfue:
Yes, it works!

PR #238, opened 2020-04-14
docs: update development/release docs

### LINKED ISSUE #235: code does not compile for wifi kit 32 any more
i wanted to flash via usb and set the arduino ide to `heltec wifi kit 32`.

the code does not compile any more due to missing LoRa pin definitions (e.g. `DIO2`).

wanted to go the usb way because i needed usb logging anyway to debug some code.

besides that, for the initial flash of a new device, one needs to compile for correct hw platform to get correct partition layout.

additionally, I managed to lose OTA capability by flashing a buggy firmware onto the device that crashes and reboots all the time (thus no WiFi OTA flashing).

thus, with the current state of device, no OTA flashing and with current state of the code, no USB flashing from arduino IDE.

ThomasWaldmann:
how to create and test binaries, IDE settings, ...

Fixes #235.

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -86,14 +86,24 @@ main repository.
 .. _Sphinx: https://www.sphinx-doc.org/
 
 
-Creating binaries
------------------
+Flashing devices / creating binaries
+------------------------------------
 
-Adjust userdefines.h, then use e.g.:
+Arduino IDE:
 
-::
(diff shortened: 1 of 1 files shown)


rexfue:
Sounds good. I would add, that if you have a new wifi kit with 8MB you can use the default_8MB partition scheme. But that can als be added later.

ThomasWaldmann:
One could, but currently it doesn't make a dfiference.

PR #245, opened 2020-05-02
slow down main loop

### LINKED ISSUE #222: how fast is the main loop() spinning?
at some time, we should determine that.

if it spins much faster than necessary, we maybe could add some sleep or do it more event-driven.

ThomasWaldmann:
fixes #222.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -55,6 +55,12 @@
 // Minimum amount of GM pulses required to early-update the display.
 #define MINCOUNTS 100
 
+// Target loop duration [ms]
+// slow down the arduino main loop so it spins about once per LOOP_DURATION -
+// this is more than enough, but still less than once per 1-2ms (without delay).
+// it should not be too slow because the speaker/LED tick is done in main loop.
+#define LOOP_DURATION 5
+
 // DIP switches
 static Switches switches;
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@rexfue ^^^ merge?

ThomasWaldmann:
rebased on current master, merge?

multigeiger/multigeiger.ino
```diff
@@ -211,7 +217,12 @@ void loop() {
     last_GMC_counts = GMC_counts;
   }
 
-  iotWebConf.doLoop();  // see webconf.cpp
+  long loop_duration;
+  loop_duration = millis() - current_ms;
+  iotWebConf.delay((loop_duration < LOOP_DURATION) ? (LOOP_DURATION - loop_duration) : 0);
+  loop_duration = millis() - current_ms;
+  log(DEBUG, "loop_duration: %ld", loop_duration);
```
rexfue, line 224: Would set it to INFO

rexfue:
OK, I tested it. See review comment

multigeiger/multigeiger.ino
ThomasWaldmann: this is nothing interesting for a normal user, thus debug is better.

rexfue: then remove it: you can't hardly see any other debug message, if this comes every 5msec

ThomasWaldmann: ok, i'll remove output and computation.

doing an INFO-level log() call would not have helped to not see it, because INFO level stuff is of course also seen when the logger is on DEBUG log level.

rexfue: OK, you're right. I always think the wrong way with debug levels :-(

PR #246, opened 2020-05-02
dip switches: only read once at boot time, fixes #207

### LINKED ISSUE #207: dip switch reading
in the current master branch code, the dip switches are read at every turn of the main loop.

is this intended / does this make sense?

looking at the hardware, one can not reach the dip switch, which sits on the pcb below the microcontroller. or shall the dip switch be put on the other side of the pcb?

ThomasWaldmann:


ThomasWaldmann:
merge?

rexfue:
OK to merge

Issue #66, opened 2019-11-17
decouple hv charging from display update interval


ThomasWaldmann:
```
  #define DISPLAYREFRESH 600000
  #define MAXCOUNTS 10000
  // Check if there are enough pulses detected or if enough time has elapsed.
  // If yes, then it is time to charge the HV capacitor and calculate the pulse rate.
  if ((GMC_counts >= MAXCOUNTS) || ((millis() - time2display) >= DISPLAYREFRESH )) {
    ...
    // charge HV capacitor
    ...
    // compute CPM
    ...
    // ... and display it.
```
The code is arranged a bit strangly - I guess it works ok as long as the DISPLAYREFRESH and MAXCOUNTS are kept about like they are, but I think logically, there are 2 issues here:

- The hv capacitor should be charged when needed and this has nothing to do with the DISPLAYREFRESH time.

- Looks like MAXCOUNTS is used for two purposes (so I guess we need MAXCOUNTRECHARGE and MAXCOUNTDISPLAY instead):

  * to trigger a display update (idea: if we have a high count, we can update more quickly without losing precision). How high that value should be for that purpose is a matter of precision requirement, see #38, so one might want to go to a higher value. (MAXCOUNTDISPLAY)
  * to trigger a hv capacitor recharge. this is independent from the other purpose and might need to be kept low to not let the capacitor discharge by too many gm pulses.  (MAXCOUNTRECHARGE)

ThomasWaldmann:
There is likely a similar issue with logging, which is also done inside that block (and thus coupled into that although it shouldn't be).

rexfue:
A view lines above of the referred lines, there you see  the charging of the cap. That happens every second. And that is independent of display or calculation time.
So I think, we should delete the two lines in the display/calulation block, that charge the capacitor:
    HV_pulse_count = jb_HV_gen_charge__chargepules();         // charge HV capacitor
    hvpulsecnt2send += HV_pulse_count;                        // count for sending
I'll do that in the next PR.
Concerning logging: I guess it should be in that block: every time, we calc a new value, we log it to serial out, why not?

ThomasWaldmann:
I didn't do voltage measurements, but I understood the code like that charging the hv capacitor is needed after some pulses because they discharge the capacitor more and more.

If the pulse rate is high, that once per second charge might not be enough maybe.

But that has nothing to do with the display update, that is my point.

[...]

Issue #38, opened 2019-11-12
measurement interval / display update interval too short


ThomasWaldmann:
Last sunday, friends and I assembled 3 geigers at shackspace.de.

When seeing them work, I noticed that there is a problem with how measurements are done / displayed. It seems that the code is rather keen on quickly displaying ....... something.

So the first value after powering up appeared only a few (N) seconds later. As it displays CPM and does not even measure a full minute for that, it has to compute cpm = counts * 60 / N.

This is bad as due to the random of radioactive decay (and maybe also other reasons in physics) and the big variation in short measurement periods for cases of low (normal) radiation, the resulting cpm output varies hugely and thus is pretty useless.

Assuming N == 6 and a real CPM value of 20 (like SBM-20 and background radiation):

```
time [s]  counts     displayed CPM
----------------------------------------------
0..6      5                     50
6..12     1                     10
12..18    2                     20
18..24    0                     0
24..30    2                     20
...
```

So the only thing more or less known after these displayed values is "ok, it's not zero and it is also not 1000 CPM, but maybe something around 20 CPM" - but you can only tell that by smoothing the values in your mind.

A better and still simple way would be to measure longer, update less frequently.

Or even better, display an average value over the last minute (sliding windows over measurement series) and do not display anything before "enough" pulses have been counted or one minute has elapsed.

ThomasWaldmann:
Or: adjust sliding window width over measurements to total counts seen in that window.

E.g. if we only get 20 CPM, let the window be 1 minute wide to have some precision.

If we get 200 CPM, the window could be 6s to get similar precision.

If we get 2000 CPM, we could have 0.6s window for similar precision.

If we get 20000 CPM, we could still have 0.6s window as we do not need to update the display any quicker than that.

rexfue:
At the moment we have 10sec Display and calculating interval. I think, that's OK.
If you switch it on for the first time and you have to wait 60sec until you see something - that is frustrating.

ThomasWaldmann:
I disagree about the "OK". But I guess it depends whether you see it as a measurement device or something for show purposes. I don't expect high precision, but I also don't expect display of values that aren't worth much. :-)

10 secs, Si22G (relatively big tube), normal german background radiation means 7 or so pulses. With a SBM20 (or other smaller tube), it would be only 2 pulses.

That can easily vary widely, just by some random pulses more or less, so what one gets is a display value that jumps arounds significantly.

[...]

PR #247, opened 2020-05-02
simplify ISR, refactor main loop

### LINKED ISSUE #220: simplify / speed up the ISR, reduce locked time
I'ld like to:
- simplify, reduce and speed up the ISR code
- simplify the code in the main loop interacting with the ISR variables
- reduce locked time by less and shorter critical sections
- do all pulse count bookkeeping in the main loop

What I think of is this:
- keep the ISR code as minimal as possible, especially remove isr_GMC_counts_2send, isr_count_timestamp_2send and isr_gotGMCpulse.
- all the different counters for display, 1-minute-log, 2.5-minute-transmission, all-time, ... would be just kept in the main loop and updated there (not in ISR).
- in the main loop, after fetching the counts from ISR, always reset the isr counter to 0 and do all other bookkeeping in main loop. this should make the code much simpler.
- enclose PIN_TEST_OUTPUT operations in some #ifdef TEST_PIN_ENABLE .. #endif

ThomasWaldmann:
fixes #220 and #66.  #38 can now trivially be addressed also.

this is a rather big and likely hard to review change, but the code needed this as the lots of bookkeeping variables and the rather lengthy main loop and strange coupling in there was not easily manageable / understandable any more. maybe it is easier to look at the resulting code rather than at the diff.

guess the ISR is now as simple and as fast as it gets, no duplication and no longer-term bookkeeping in there any more. also no debugging overhead when not needed.

the main code does not interact with the ISR variables any more, there is just one call to fetch the interesting values (which also resets the in-ISR counter). so all the mutex stuff is now in the tube module.

the main loop was dissected into functions and the state-keeping static variables are now local to the respective function and thus easier to understand. because the variables now live in separate namespaces, a lot of variable names could be made shorter / easier.

there are only few variables left in the main loop: the always increasing counters for GM counts and HV pulses, the last GM count timestamp and the time between the last GM counts.

having these "master counters" in this "always-increasing mode" and the bookkeeping local in the misc. functions (just remembering the "last" values when they did something) makes it easy for the function to compute the deltas for the counts or timestamps for whatever interval needed.

Tested by me:
- switching from info display to "zero display" to "measurement display" works
- displayed values seem correct
- serial/USB log output seems correct

Not tested:
- data transmission to internet servers


rexfue:
OK, I'll check tomorrow

rexfue:
Checked it: sends to servers and also to LoRa. So ist OK to merge.

PR #251, opened 2020-05-04
timer-driven sequencer, hw PWM sound, fixes #35

### LINKED ISSUE #35: use pwm for sound?
There is some code toggling GPIOs by software to make the tick sound and some startup sound.

The ESP32 has PWM hardware that can be programmed with frequency, ratio and then runs completely in hardware without using the CPU for toggling pins or doing delays.

Currently, the piezo is connected to 2 GPIOs that are driven inverted to each other to cleverly double the amplitude (by going from hi/lo to lo/hi). Is that needed to be loud enough or would half of that also suffice?

Not sure if we could easily have 2 PWMs that are offset so that they come out inverted.

Note: we could use single PWM even with the current PCB, wasting 1 IO pin by just pulling it low permanently while the other is doing PWM.

ThomasWaldmann:
hardware PWM for sound (ticks and tones).

this is now a hw timer interrupt driven sequencer which uses hw PWM for sound output. :-)

high volume: drives the PWM outputs in a complementary way
low volume: one PWM output, other output fixed low.

thus, output power (== volume) should be as before.

the high-pitch geiger tick is now directly initiated from ISR code.
there is also the capability for a low-pitch "tock" now, can be used for debugging.

the main loop only spins once a second now.

### DIFF
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -33,7 +33,7 @@ void IRAM_ATTR isr_audio() {
   // exceptions like "Cache disabled but cached memory region accessed".
   static unsigned int current = 0;  // current period counter
   static unsigned int next = PERIODS(1000);  // periods to next sequencer execution
-  if (current++ < next)
+  if (++current < next)
     return;  // nothing to do yet
 
   // we reached "next", so we execute the sequencer:
@@ -54,8 +54,8 @@ void IRAM_ATTR isr_audio() {
       playing_tick = true;
     }
   }
-  volatile int *p = isr_sequence;
-  if (p) {
+  if (isr_sequence) {
+    volatile int *p = isr_sequence;
     frequency_mHz = *p++;
     volume = *p++;
     led = *p++;
@@ -64,7 +64,7 @@ void IRAM_ATTR isr_audio() {
   }
   portEXIT_CRITICAL_ISR(&mux_audio);
 
-  if (!p)
+  if (!isr_sequence)
     return;  // nothing to do
 
   // note: by all means, **AVOID** mcpwm_set_duty() in ISR, causes floating point coprocessor troubles!
@@ -177,6 +177,7 @@ void setup_speaker(bool playSound, bool use_led, bool use_speaker) {
       TONE(1046502, 1, -1, 2),  // C
       TONE(987767, 1, -1, 4),   // H
       TONE(987767, 0, -1, 4),   // H
+      TONE(0, 0, 0, 2),         // ---
       TONE(0, 0, -1, 0),        // speaker off, end
     };
     play((int *)sequence);
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -2,83 +2,183 @@
 // also handles the onboard LED, which lights up while speaker ticks.
 
 #include <Arduino.h>
+#include <driver/mcpwm.h>
 
 #include "speaker.h"
+#include "timers.h"
 
 #define PIN_SPEAKER_OUTPUT_P 12
 #define PIN_SPEAKER_OUTPUT_N 0
 
-void setup_speaker() {
+// MUX (mutexes used for mutual exclusive access to isr variables)
+portMUX_TYPE mux_audio = portMUX_INITIALIZER_UNLOCKED;
+
+volatile int *isr_audio_sequence = NULL;
+volatile int *isr_tick_sequence = NULL;
+volatile int *isr_sequence = NULL;  // currently played sequence
+
+static int tick_sequence[8], tock_sequence[8];
+
+// hw timer period and microseconds -> periods conversion
+#define PERIOD_DURATION_US 1000
+#define PERIODS(us) ((us) / PERIOD_DURATION_US)
+
+void IRAM_ATTR isr_audio() {
+  // this code is periodically called by a timer hw interrupt, always same period.
+  // we need to decide internally whether we actually want to do something.
+  //
+  // note: this is implemented like it is because dynamically reprogramming the hw timer
+  // to a different period would require us to call library functions like timerAlarmWrite
+  // which are **not** in IRAM (but in flash) and doing that can lead to spurious fatal
+  // exceptions like "Cache disabled but cached memory region accessed".
+  static unsigned int current = 0;  // current period counter
+  static unsigned int next = PERIODS(1000);  // periods to next sequencer execution
+  if (current++ < next)
+    return;  // nothing to do yet
+
+  // we reached "next", so we execute the sequencer:
+  current = 0;
+
+  // tone and tick generation, also led blinking
+  int frequency_mHz, volume, led, duration_ms;
+  static bool playing_audio = false, playing_tick = false;
+
+  // fetch next tone / next led state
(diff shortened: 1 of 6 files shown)


multigeiger/speaker.cpp
```diff
@@ -31,34 +52,43 @@ void cycle(int t1, int t2, int volume) {
...
-}
-
-void tone(int frequency_mHz, int time_ms, int volume) {
+void tone(int frequency_mHz, int duration_ms, int volume) {
+  // volume: 0: low, 1: high
+  #if PWM
+  mcpwm_set_frequency(MCPWM_UNIT_0, MCPWM_TIMER_0, frequency_mHz / 1000);
+  // enable sound output
+  float duty;
+  duty = volume ? 50.0 : 10.0;
```
ThomasWaldmann, line 61: maybe not the best way to do this, last tone of melody sounds a bit different than previously.

multigeiger/speaker.cpp
```diff
@@ -31,34 +52,43 @@ void cycle(int t1, int t2, int volume) {
...
+  mcpwm_set_duty(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_A, duty);
+  mcpwm_set_duty(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_B, duty);
+  // wait
+  delay(duration_ms);
```
ThomasWaldmann, line 65: note that this is completely idle cpu (compare to sw version).

ThomasWaldmann:
```
E (843) MCPWM: /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/driver/mcpwm.c:190 (mcpwm_set_duty_type):MCPWM DUTY TYPE ERROR
E (856) MCPWM: /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/driver/mcpwm.c:190 (mcpwm_set_duty_type):MCPWM DUTY TYPE ERROR
```

Seen once directly after boot. I don't see an error in the calls.

ThomasWaldmann:
rebased on current master, collapsed changesets into one.

ThomasWaldmann:
rebased on current master.

ThomasWaldmann:
merge?

rexfue:
The start sound sounds ugly!
With speaker tick, there are two different impulse series: one with 10 time 500usec and another with 48 times 100usec. Is that, what you want?

rexfue:
Only for information (i don't think, its important), the two pulse streams don't start at the same time:
ScreenImg

ThomasWaldmann:
@rexfue oh, interesting. it doesn't matter for the tones with longer duration for sure, it matters a little for the ticks, which are only few periods of the signal.

multigeiger/speaker.cpp
ThomasWaldmann: @rexfue this is why the startup sound sounds a bit ugly at the end, because the last note has a duty cycle of 10% (not 50% as all others).

I guess I'll try changing it to 50% also and hear if it gets better then.

ThomasWaldmann:
100us * 2 = 200us -> 5kHz "tick" == geiger event

500us * 2 = 1000us -> 1kHz "tock" == capacitor fully charged event

rexfue:
Ah, I see. Sounds like good idea. But do you think, that is what the normal user expects? On every Geiger the tick is only linked to the count pulse, nothing else.

ThomasWaldmann:
While debugging, I was interested in a "realtime" feedback from charging, thus i added this "tock".

I can remove the call (so end users won't get confused), but i'ld keep the functionality inside `tick()`, so it can be used by developers.

rexfue:
ok, that could be a good solution.

multigeiger/speaker.cpp
ThomasWaldmann: at 50% it sounds ok.

ThomasWaldmann:
OK, guess it is ready to merge?

ThomasWaldmann:
I implemented "volume" and got rid of the floating point inside the ISR.

Now it should be quite similar to the previous sw implementation.

multigeiger/speaker.cpp
```diff
@@ -147,18 +153,12 @@ void setup_speaker(bool playSound, bool use_led, bool use_speaker) {
...
+  pwm_config.cmpr_b = 50.0;
   pwm_config.counter_mode = MCPWM_UP_COUNTER;
   mcpwm_init(MCPWM_UNIT_0, MCPWM_TIMER_0, &pwm_config);
 
-  // MCPWM A/B outputs shall be inverted (more power to speaker):
-  mcpwm_set_duty_type(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_A, MCPWM_DUTY_MODE_0);
-  mcpwm_set_duty_type(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_B, MCPWM_DUTY_MODE_1);
-
-  mcpwm_start(MCPWM_UNIT_0, MCPWM_TIMER_0);
-
```
ThomasWaldmann: this is done in ISR anyway.

rexfue:
The only left problem is tone nbr. 3, it is too short. Tone 1 and 2 are ca. 170msec long, tone 3 is only ca. 100msec long. And the pause between 2 and 3 is too long (203 msec instead of 170 msec). In the old version all 3 tones were ca. 170ms long and all pauses were ca. 170ms.
Is sounds not really good, but i guess, it can be merged anyway.

multigeiger/speaker.cpp
```diff
@@ -176,7 +176,7 @@ void setup_speaker(bool playSound, bool use_led, bool use_speaker) {
       TONE(987767, 1, -1, 2),   // H
       TONE(1046502, 1, -1, 2),  // C
       TONE(987767, 1, -1, 4),   // H
-      TONE(987767, 1, -1, 4),   // H
+      TONE(987767, 0, -1, 4),   // H
```
ThomasWaldmann: 0 was the original volume "as before".

it now works without sounding ugly.

rexfue:
OK, sounds good!

ThomasWaldmann:
The tone/pause duration differences are reproducible?

They have the same "duration" value and the timer interval is 1ms, so there should be no that big differences.

rexfue:
Yes, was in the last version (before implementing volume). Now its OK.

ThomasWaldmann:
Hmm, i didn't change how timing works. Maybe a side effect of the calls I removed.

multigeiger/speaker.cpp
```diff
@@ -33,7 +33,7 @@ void IRAM_ATTR isr_audio() {
   // exceptions like "Cache disabled but cached memory region accessed".
   static unsigned int current = 0;  // current period counter
   static unsigned int next = PERIODS(1000);  // periods to next sequencer execution
-  if (current++ < next)
+  if (++current < next)
```
ThomasWaldmann: if it wants a 1 ms interval, `PERIODS(1000)` will result in `1` meaning that the ISR needs to run every time it is called.

it was run every second time before that fix.

PR #253, opened 2020-05-05
display.cpp - new font for status, prepare BLE connectivity, CPS->CPM

t-pi:
update display.cpp 
- new font for status line (u8x8 uses 8px character width anyway), 
- prepare for BLE connectivity, 
- changed symbol CPS to CPM

### DIFF
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -110,7 +110,7 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool conne
 
     // print the upper line including time and measured radation
     pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
-    pu8x8->print(ble_active ? '\x4A' : '\x20');
+    pu8x8->print(ble_active ? '\x4A' : ' '); // 0x4A corresponds to Bluetooth symbol in selected font.
     pu8x8->setFont(u8x8_font_7x14_1x2_f);
 
     if (TimeSec < 60)    {    // < 1 minute -> display in seconds
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -112,7 +112,6 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool conne
     pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
     sprintf(output, "%c", 0x4A);
     if (ble_active) pu8x8->print(output);
-    
     pu8x8->setFont(u8x8_font_7x14_1x2_f);
 
     if (TimeSec < 60)    {    // < 1 minute -> display in seconds
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -74,7 +74,7 @@ void clearDisplayLine(int line) {
 
 void displayStatusLine(String txt) {
   int line = isLoraBoard ? 5 : 7;
-  pu8x8->setFont(u8x8_font_5x8_f);
+  pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
   clearDisplayLine(line);
   pu8x8->drawString(0, line, txt.c_str());
 }
@@ -90,7 +90,7 @@ char *nullFill(int n, int digits) {
   return erg;
 }
 
-void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool connected) {
+void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool connected, bool ble_active) {
   if (!use_display) {
     if (!displayIsClear) {
       pu8x8->clear();
@@ -106,28 +106,34 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
   if (!isLoraBoard) {
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
-    if (TimeMin >= 999) TimeMin = 999;  // limit minutes to max. 999
+    TimeMin = TimeMin % 999;  // limit minutes to max. 999, roll over
 
     // print the upper line including time and measured radation
+    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
+    sprintf(output, "%c", 0x4A);
+    if (ble_active) pu8x8->print(output);
+    
     pu8x8->setFont(u8x8_font_7x14_1x2_f);
 
-    if (TimeMin >= 1) {                 // >= 1 minute -> display in minutes
+    if (TimeSec < 60)    {    // < 1 minute -> display in seconds
+      sprintf(output, "%2ds", TimeSec);
+      pu8x8->print(output);
+    } else {                  // >= 1 minute -> display in minutes
       sprintf(output, "%3d", TimeMin);
       pu8x8->print(output);
-    } else {                            // < 1 minute -> display in seconds, inverse
-      sprintf(output, "%3d", TimeSec);
-      pu8x8->inverse();
-      pu8x8->print(output);
-      pu8x8->noInverse();
     }
-    sprintf(output, "%7d nSv/h", RadNSvph);
+    if (ble_active) {
+      sprintf(output, "%6d nSv/h", RadNSvph);
+    } else {
+      sprintf(output, "%7d nSv/h", RadNSvph);
+    }
     pu8x8->print(output);
     pu8x8->setFont(u8x8_font_inb33_3x6_n);
-    pu8x8->drawString(0, 2, nullFill(CPS, 5));
+    pu8x8->drawString(0, 2, nullFill(CPM, 5));
   } else {
     pu8x8->setFont(u8x8_font_5x8_f);
     pu8x8->drawString(0, 2, nullFill(RadNSvph, 8));
-    pu8x8->draw2x2String(0, 3, nullFill(CPS, 4));
+    pu8x8->draw2x2String(0, 3, nullFill(CPM, 4));
     pu8x8->drawString(0, 5, "     cpm");
   }
   displayStatusLine(connected ? " " : "connecting...");
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -106,12 +106,11 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool conne
   if (!isLoraBoard) {
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
-    TimeMin = TimeMin % 999;  // limit minutes to max. 999, roll over
+    TimeMin = TimeMin % 1000;  // limit minutes to max. 999, roll over
 
     // print the upper line including time and measured radation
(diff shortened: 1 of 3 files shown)


t-pi:
first time pull request, any hints are welcome :)

multigeiger/display.cpp
```diff
@@ -74,7 +74,7 @@ void clearDisplayLine(int line) {
 
 void displayStatusLine(String txt) {
   int line = isLoraBoard ? 5 : 7;
-  pu8x8->setFont(u8x8_font_5x8_f);
+  pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
```
ThomasWaldmann, line 77: hah, "amstrad cpc" :)

multigeiger/display.cpp
```diff
@@ -106,28 +106,33 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
   if (!isLoraBoard) {
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
-    if (TimeMin >= 999) TimeMin = 999;  // limit minutes to max. 999
+    TimeMin = TimeMin % 999;  // limit minutes to max. 999, roll over
```
ThomasWaldmann, line 109: ah, good idea! 

but: off by one. you need to do % 1000 to get 999 max.

multigeiger/display.cpp
```diff
@@ -106,28 +106,33 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
...
     // print the upper line including time and measured radation
+    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
+    sprintf(output, "%c", 0x4A);
+    if (ble_active) pu8x8->print(output);
     pu8x8->setFont(u8x8_font_7x14_1x2_f);
 
-    if (TimeMin >= 1) {                 // >= 1 minute -> display in minutes
-      sprintf(output, "%3d", TimeMin);
+    if (TimeSec < 60)    {    // < 1 minute -> display in seconds
+      sprintf(output, "%2ds", TimeSec);
```
ThomasWaldmann, line 118: good idea. if we have 3 digits for the minutes, 2 digits + "s" for the seconds will also fit.

multigeiger/display.cpp
```diff
@@ -106,28 +106,33 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
...
+      sprintf(output, "%3d", TimeMin);
       pu8x8->print(output);
-      pu8x8->noInverse();
     }
-    sprintf(output, "%7d nSv/h", RadNSvph);
+    if (ble_active) {
+      sprintf(output, "%6d nSv/h", RadNSvph);
+    } else {
+      sprintf(output, "%7d nSv/h", RadNSvph);
+    }
```
ThomasWaldmann, line 128: see "nothing shifts around" comment above.

multigeiger/display.cpp
```diff
@@ -90,7 +90,7 @@ char *nullFill(int n, int digits) {
   return erg;
 }
 
-void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool connected) {
+void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool connected, bool ble_active) {
```
ThomasWaldmann, line 93: oh, good catch.

multigeiger/display.h
```diff
@@ -4,7 +4,7 @@
 #define _DISPLAY_H_
 
 void setup_display(bool loraHardware);
-void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool connected);
+void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool connected, bool ble_active);
```
ThomasWaldmann, line 7: CPM

multigeiger/multigeiger.ino
```diff
@@ -117,7 +117,7 @@ void display(unsigned long current_ms, unsigned long current_counts, unsigned lo
 
     // ... and display them.
     DisplayGMC(((int)accumulated_time / 1000), (int)(accumulated_Dose_Rate * 1000), (int)(Count_Rate * 60),
-               (showDisplay && switches.display_on), wifi_connected);
+               (showDisplay && switches.display_on), wifi_connected, false); // last false in preparation for future BLE connectivity
```
ThomasWaldmann, line 120: maybe add ble_connected or ble_active as variable to main loop (and put the comment there), and use the variable here?

multigeiger/multigeiger.ino
```diff
@@ -130,7 +130,7 @@ void display(unsigned long current_ms, unsigned long current_counts, unsigned lo
     if (afterStartTime && ((current_ms - boot_timestamp) >= afterStartTime)) {
       afterStartTime = 0;
       DisplayGMC(0, 0, 0,
-                 (showDisplay && switches.display_on), wifi_connected);
+                 (showDisplay && switches.display_on), wifi_connected, false); // last false in preparation for future BLE connectivity
```
ThomasWaldmann, line 133: maybe add ble_connected or ble_active as variable to main loop (and put the comment there), and use the variable here?

multigeiger/display.cpp
```diff
@@ -106,28 +106,33 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
   if (!isLoraBoard) {
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
-    if (TimeMin >= 999) TimeMin = 999;  // limit minutes to max. 999
+    TimeMin = TimeMin % 999;  // limit minutes to max. 999, roll over
 
     // print the upper line including time and measured radation
+    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
+    sprintf(output, "%c", 0x4A);
+    if (ble_active) pu8x8->print(output);
```
ThomasWaldmann, line 114: how about removing the sprintf and the if and using:

```
pu8x8->print(ble_active ? '\x4A' : '\x20')
```

That would output 1 character in both cases so that the rest of the layout stays the same and nothing shifts around.

multigeiger/display.cpp
ThomasWaldmann: also, a good rule is to always remember the `cond ? if_true : if_false` way to write a term when one wants to write an otherwise duplicate if/else block.

multigeiger/multigeiger.ino
ThomasWaldmann: or do not yet add the param / variable at all.

multigeiger/display.cpp
t-pi: uppsi, right you are! I'll update the pull request with all changes soon...

multigeiger/display.cpp
t-pi: felt more natural :)
Especially as I have wondered a long time what that counter represents. I hope the "s" will give it away more clearly...

multigeiger/display.cpp
t-pi: I was not so sure whether that works with two full commands... will adapt.

multigeiger/display.h
t-pi: good catch ;)

multigeiger/multigeiger.ino
t-pi: I need the variable to add the BLE symbol code. Will add an always-false variable in multigeiger.ino

multigeiger/display.cpp
t-pi: That was actually my first version, then I tended towards giving more space for high counts if needed. If the shifting is not wanted, I'll revert to that first version.

t-pi:
actually, BLE connectivity is already working locally for several days while I'm working on a simple app prototype. 
This pull request is a small first step to test and hone my git / github mojo 🤓

multigeiger/display.cpp
t-pi: Actually, reverted it to single %6d line to avoid shifting text.

multigeiger/display.cpp
t-pi: some things never die ;)

multigeiger/display.cpp
```diff
@@ -106,28 +106,28 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPS, bool use_display, bool conne
   if (!isLoraBoard) {
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
-    if (TimeMin >= 999) TimeMin = 999;  // limit minutes to max. 999
+    TimeMin = TimeMin % 1000;  // limit minutes to max. 999, roll over
 
     // print the upper line including time and measured radation
+    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
+    pu8x8->print(ble_active ? '\x4A' : '\x20');
```
ThomasWaldmann, line 113: 4A is 'J'? If so, we could also write 'J' there.

multigeiger/multigeiger.ino
```diff
@@ -117,7 +118,7 @@ void display(unsigned long current_ms, unsigned long current_counts, unsigned lo
 
     // ... and display them.
     DisplayGMC(((int)accumulated_time / 1000), (int)(accumulated_Dose_Rate * 1000), (int)(Count_Rate * 60),
-               (showDisplay && switches.display_on), wifi_connected);
+               (showDisplay && switches.display_on), wifi_connected, ble_connected); // ble_connected in preparation for future BLE connectivity
```
ThomasWaldmann, line 121: maybe move the comment to line 95.

multigeiger/multigeiger.ino
```diff
@@ -130,7 +131,7 @@ void display(unsigned long current_ms, unsigned long current_counts, unsigned lo
     if (afterStartTime && ((current_ms - boot_timestamp) >= afterStartTime)) {
       afterStartTime = 0;
       DisplayGMC(0, 0, 0,
-                 (showDisplay && switches.display_on), wifi_connected);
+                 (showDisplay && switches.display_on), wifi_connected, ble_connected); // ble_connected in preparation for future BLE connectivity
```
ThomasWaldmann, line 134: maybe move the comment to line 95.

multigeiger/display.cpp
t-pi: Not in the selected font, there it represents a Bluetooth symbol. I've added a comment and changed '\x20' to ' '

multigeiger/multigeiger.ino
t-pi: good proposition, done.

multigeiger/multigeiger.ino
t-pi: s. above, done.

ThomasWaldmann:
@t-pi on what kind of device did you test your changes?

t-pi:
I have here a Wifi Stick 32. Should I test also on Wireless Stick with Lora? Basically I did not touch any LoraStick related code.

ThomasWaldmann:
Ah, ok, if you're sure about that, I guess it is not needed.

PR #254, opened 2020-05-07
timer-driven adaptive recharging

### LINKED ISSUE #192: HV pulse every sec using a timer
It seems, that if Wifi connection has problems, the HV will not be charged for many seconds.
So we should use a hardware timer and timer interrupt to charge HV every second.

ThomasWaldmann:
This does all HV charging via a timer driven ISR.

This fixes #192, but in an improved way:

Just doing a HV charge pulse "every second" is not good enough:
- if radiation-caused discharging and discharging caused by other leaks are low, a pulse every second is more than needed. No big problem, just drawing more power than needed.
- if radiation or leaks are high, doing only 1 pulse per second is not enough, so we need to adapt to not let the HV voltage break down.

The code adapts to this by targeting a 2-pulse recharge.

Due to the way we charge, we always will have 1 charge pulse - no matter whether we actually needed to charge or not. But 2 pulses are only needed if we really needed to charge. Thus the code targets doing 2 pulses.

If less than 2 pulses are needed, we are maybe charging too often - the code will then slightly increase the time between charge cycles.

If 2 pulses are needed, do not change the timer interval.

If more than 2 pulses are needed, we are charging too infrequently, decrease the time between charge cycles. This is done proportional to `1 / charge_pulses_needed`. So, if we encounter high radiation, have a high discharge and need many recharge pulses, we will quickly adapt to recharging way more frequently.

If HV charging fails, the code sets a `hv_error` variable in main loop now, but it is not used there yet.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -218,6 +218,7 @@ void tick_blink(unsigned long current_counts) {
 
 void loop() {
   static bool wifi_connected = false;
+  static bool hv_error = false;  // TODO: display this somewhere
 
   static bool have_thp = false;
   static float temperature = 0.0, humidity = 0.0, pressure = 0.0;
@@ -247,7 +248,7 @@ void loop() {
 
   read_THP(current_ms, &have_thp, &temperature, &humidity, &pressure);
 
-  hv_pulses += charge_hv(gm_counts, current_ms);
+  read_hv(&hv_error, &hv_pulses);
 
   display(current_ms, gm_counts, gm_count_timestamp, hv_pulses, wifi_connected);
 
--- a/multigeiger/timers.cpp
+++ b/multigeiger/timers.cpp
@@ -1,0 +1,12 @@
+#include <Arduino.h>
+
+#define RECHARGE_TIMER 0
+hw_timer_t *recharge_timer = NULL;
+
+void setup_recharge_timer(void (*isr_recharge)(), int period_us) {
+  recharge_timer = timerBegin(RECHARGE_TIMER, 80, true);  // prescaler: 80MHz / 80 == 1MHz
+  timerAttachInterrupt(recharge_timer, isr_recharge, true);  // set ISR
+  timerAlarmWrite(recharge_timer, period_us, true);  // set alarm after period, do repeat
+  timerWrite(recharge_timer, 0);
+  timerAlarmEnable(recharge_timer);
+}
--- a/multigeiger/tube.cpp
+++ b/multigeiger/tube.cpp
@@ -5,6 +5,7 @@
 #include <Arduino.h>
 
 #include "log.h"
+#include "timers.h"
 #include "tube.h"
 
 // The test pin, if enabled, is high while isr_GMC_count is active.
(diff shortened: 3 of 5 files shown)


multigeiger/tube.cpp
```diff
@@ -35,22 +35,121 @@ TUBETYPE tubes[] = {
...
+
+// How many HV capacitor charge pulses to generate for any later charge cycle.
+#define MAX_CHARGE_PULSES 333
+
+// How many HV capacitor charge pulses to generate for first charge.
+// After device power on, HV capacitor is completely empty.
+#define MAX_CHARGE_PULSES_INITIAL 3333
+
+void IRAM_ATTR isr_recharge() {
+  static int state = 0;
```
ThomasWaldmann, line 66: i implemented a state machine here to keep the in-ISR time low.

the old code just used `delayMicroseconds()`, but we should not do that in a ISR.

multigeiger/tube.cpp
```diff
@@ -35,22 +35,121 @@ TUBETYPE tubes[] = {
...
+      if (charge_pulses <= 1) {
+        // one charge pulse was enough, so likely we charge too often
+	next = next * 5 / 4;
+      } else {
+	// 2 charge pulses: no change
+        // more: we charge too rarely, adapt in a linear way
+	next = next * 2 / charge_pulses;
+      }
+    } 
+    // next = min(max(next, 1000), 10 * HVPULSE_MS * 1000)  // bounds: 1ms .. 10*HVPULSE_MS ms
```
ThomasWaldmann, line 125: if we have a reason to have lower and upper bounds, we could use code like this.

if we do not find a reason, i'll remove this.

multigeiger/tube.cpp
```diff
@@ -83,50 +182,6 @@ void read_GMC(unsigned long *counts, unsigned long *timestamp, unsigned int *bet
...
-
-// How many HV capacitor charge pulses to generate for first charge,
-// for the call in setup(). Not timing critical and capacitor is empty.
-#define MAX_CHARGE_PULSES_INITIAL 3333
-
-// Interval for unconditional HV charge pulse generation. [msec]
-#define HVPULSE_MS 1000
-
-// After this many GM counts, we want to recharge the HV capacitor, too.
-#define HVPULSE_COUNTS 100
```
ThomasWaldmann, line 99: this is done way more flexible now.

multigeiger/tube.cpp
```diff
@@ -35,22 +35,121 @@ TUBETYPE tubes[] = {
...
+  timerWrite(timer, 0);
+  timerAlarmEnable(timer);
+}
+
+// How many HV capacitor charge pulses to generate for any later charge cycle.
+#define MAX_CHARGE_PULSES 333
+
+// How many HV capacitor charge pulses to generate for first charge.
+// After device power on, HV capacitor is completely empty.
+#define MAX_CHARGE_PULSES_INITIAL 3333
```
ThomasWaldmann, line 63: I guess we could also simplify the code and use the higher value always.

The code will now stop doing recharges anyway if that amount of pulses does not fully charge the capacitor.

multigeiger/tube.cpp
ThomasWaldmann: note: this is pseudo code, min/max seems to not exist.

ThomasWaldmann:
@rexfue @WwwBoehriDe have a look, please!

multigeiger/tube.cpp
t-pi: Could you please give a short comment in the code what the states represent? thx

multigeiger/tube.cpp
t-pi: As the charging stops once full anyway, a single value is enough IMO. 
Worst case during operation? So many pulse that the HV capacitor is empty again and must be recharged.
Lower value only required if we want to react more quickly with alert / warning during operation. Is that the case?

multigeiger/tube.cpp
```diff
@@ -35,22 +35,121 @@ TUBETYPE tubes[] = {
...
+    } else {
+      // depending on a lot of circumstances (e.g. level of radiation, humidity,
+      // leak currents, tube type, ...), we might need to charge the HV capacitor
+      // more or less often. we target charging with 2 pulses here.
+      if (charge_pulses <= 1) {
+        // one charge pulse was enough, so likely we charge too often
+        next = next * 5 / 4;
+      } else {
+        // 2 charge pulses: no change
+        // more: we charge too rarely, adapt in a linear way
```
t-pi, line 121: not really linear... more like hyperbolic =)

multigeiger/tube.cpp
t-pi: feels healthier to me to have some bounds. Maybe we can start with 1 ms .. 10 s and test with some yellow glowing objects?

multigeiger/tube.cpp
```diff
@@ -35,22 +35,121 @@ TUBETYPE tubes[] = {
...
+        next = next * 2 / charge_pulses;
+      }
+    }
+    // next = min(max(next, 1000), 10 * HVPULSE_MS * 1000)  // bounds: 1ms .. 10*HVPULSE_MS ms
+    reschedule(recharge_timer, next);
+    return;
+  }
+  if (state == 5) {
+    // capacitor does not charge! give up -> do not reschedule.
```
t-pi, line 130: Is this handled somewhere already with e.g. mandatory display on, warning sound etc.?
Otherwise we might leave the user baffled why it isn't working anymore. So maybe some reschedule in the far future (1 min? 10 min?) might be still helpful?

multigeiger/tube.cpp
ThomasWaldmann: 10 times as many charge pulses will cause 10 times more frequent charging.

multigeiger/multigeiger.ino
```diff
@@ -235,7 +236,7 @@ void loop() {
 
   read_THP(current_ms, &have_thp, &temperature, &humidity, &pressure);
 
-  hv_pulses += charge_hv(gm_counts, current_ms);
+  read_hv(&hv_error, &hv_pulses);
```
ThomasWaldmann, line 239: we have the hv_error state here, but it is not displayed yet.

multigeiger/tube.cpp
ThomasWaldmann: guess I'll use an enum to make it clearer.

multigeiger/tube.cpp
ThomasWaldmann: there is only hv_error variable yet in main loop.

multigeiger/tube.cpp
ThomasWaldmann: Guess I'll just use one relatively high value.

multigeiger/tube.cpp
ThomasWaldmann: yeah, guess going below 1ms would not make sense as 1 charge pulse already takes some ms.

multigeiger/tube.cpp
t-pi: ok, indeed linear for pulses

ThomasWaldmann:
crap, it looks like there are problems calling library functions because they are in flash, not in IRAM.

mostly, it works, but not always...

so, guess I'll rewrite this so that it doesn't need timer reprogramming lib functions...

ThomasWaldmann:
OK, took a bit, but I guess I have it right now.

@t-pi Feedback was resolved.

multigeiger/tube.cpp
```diff
@@ -35,22 +36,133 @@ TUBETYPE tubes[] = {
...
+#define PERIODS(us) ((us) / PERIOD_DURATION_US)
+
+void IRAM_ATTR isr_recharge() {
+  // this code is periodically called by a timer hw interrupt, always same period.
+  // we need to decide internally whether we actually want to do something.
+  //
+  // note: this is implemented like it is because dynamically reprogramming the hw timer
+  // to a different period would require us to call library functions like timerAlarmWrite
+  // which are **not** in IRAM (but in flash) and doing that can lead to spurious fatal
+  // exceptions like "Cache disabled but cached memory region accessed".
```
ThomasWaldmann, line 65: looks like this is the way to have reliable timer ISR code on ESP32.

multigeiger/tube.cpp
```diff
@@ -35,22 +36,133 @@ TUBETYPE tubes[] = {
...
+  if (state == charge_fail) {
+    // capacitor does not charge!
+    portENTER_CRITICAL_ISR(&mux_hv);
+    isr_hv_charge_error = true;
+    isr_hv_pulses += charge_pulses;
+    portEXIT_CRITICAL_ISR(&mux_hv);
+    // let's retry charging later
+    state = init;
+    next_charge = PERIODS(1000000);  // reset to default 1s charge interval
+    next_state = PERIODS(10 * 60 * 1000000);  // wait for 10 minutes before retrying
```
ThomasWaldmann, line 148: now with "retrying later".

multigeiger/timers.cpp
```diff
@@ -0,0 +1,12 @@
...
+#define RECHARGE_TIMER 0
+hw_timer_t *recharge_timer = NULL;
+
+void setup_recharge_timer(void (*isr_recharge)(), int period_us) {
+  recharge_timer = timerBegin(RECHARGE_TIMER, 80, true);  // prescaler: 80MHz / 80 == 1MHz
+  timerAttachInterrupt(recharge_timer, isr_recharge, true);  // set ISR
+  timerAlarmWrite(recharge_timer, period_us, true);  // set alarm after period, do repeat
+  timerWrite(recharge_timer, 0);
+  timerAlarmEnable(recharge_timer);
+}
```
ThomasWaldmann, line 12: put timer stuff into own module as we'll have more of them soon, so we can avoid accidentally using the same hw timer number accidentally for different purposes.

multigeiger/tube.cpp
ThomasWaldmann: 100us granularity so we could adjust these 1500us high 1000us low timings, if needed.

it's important to neither go way too fast here (too much interrupt overhead), nor too slow (to coarse granularity).

ThomasWaldmann:
merge?

ThomasWaldmann:
@t-pi @rexfue shall i merge or do you want to review / try first?

rexfue:
Please wait til tomorrow, I want to test it and have a look on the oscilloscope - OK?

t-pi:
Will test as soon as you merge. My Lora stick has issued, counting no more after some hours. After a reset everything's fine. Would like to test with new timing and HV error displayed...

multigeiger/tube.cpp
t-pi: Look good to me!

rexfue:
It's working great.
Is it okay that sometimes there's only one pulse? 
Otherwise the controller works. With quiescent radiation it slowly goes down to 10sec priode, if I bring a piece of potsherd near by, it goes down to just over 1 sec. That looks good.
OK to merge.

ThomasWaldmann:
@rexfue yeah, sometimes only one pulse is expected, because it will only do as many pulses as needed to get the "cap full" interrupt.

so, if 1 is enough, it will only do 1.

but, if it needed only 1 pulse, there is the suspicion it is maybe charging too fast, so it is increasing the time interval between charge cycles by 25%. So, sooner or later, it'll need more than 1 pulse.

having 2 pules per charge cycle is the goal because with 2, we **know** that the capacitor really needed (a bit of) charging.

thus, a bit of oscillation considering the amount of needed pulses per charge cycle is expected. but as long as we stay well inside the plateau voltage of the tube, this should be ok.

charging too often does no harm except higher power consumption, thus we only adapt slowly to less frequent charging cycles.

OTOH, charging too rarely could lead to malfunction due to undervoltage, thus we adapt quickly if we suddenly notice we need many pulses per charge cycle.

ThomasWaldmann:
btw, please note that we currently can not easily see the amount of pulses in **one** charge cycle (except maybe on an oscilloscope).

what the code outputs is the overall amount of pulses within on display update interval or one data transmission interval.

but with my audio PWM changes, you can hear the charge cycles. :-)

Issue #202, opened 2020-02-17
use audio or upstream signalling for major errors?


ThomasWaldmann:
related to #201.

the problem is here that after the multigeiger is (finally) deployed to the outside, nobody will look at the **display** usually and only very rarely at the **ESP32 web interface**.

**audio** would work slightly better, a beeping device would get some attention (of course should be only used for serious errors).

I know this from UPS devices and RAID controllers. Sometimes only the thing beeping successfully draws attention (if the battery is dead or disks have failed).

**online** upstream notification would also work, **IF** an email notify is triggered server side ("please look after your multigeiger device ID: ESPxxxxxxxxxx").

ThomasWaldmann:
also related: #257

Issue #201, opened 2020-02-17
maintain and display device status


ThomasWaldmann:
besides the radiation status, we should maintain some device status internally, like:
- is the high voltage generation working as expected?
- wifi signal strength
- lora signal strength
- does uploading data to each configured server work?

that status should be made visible via:
- OLED display
- ESP32 web interface

ThomasWaldmann:
related to #257.

ThomasWaldmann:
#258 slightly improved this, it displays on OLED:
- HV status, ok or error
- WiFi connected or not

#265 improved so that http errors also get displayed on OLED display.

ThomasWaldmann:
moved the open stuff to #307, so this one can be closed.

PR #258, opened 2020-05-10
implement status display

### LINKED ISSUE #257: display status
maybe the OLED display could briefly show:
- wifi status (off, connecting, connected)
- ble status (off, ...)
- high voltage status (ok, error)
- data transmission status (ok, error)

anything else?

ThomasWaldmann:
fixes #257. 

slightly improves on #201 and #202.

### DIFF
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -79,6 +79,54 @@ void displayStatusLine(String txt) {
   pu8x8->drawString(0, line, txt.c_str());
 }
 
+static int status[STATUS_MAX] = {ST_NODISPLAY, ST_NODISPLAY, ST_NODISPLAY, ST_NODISPLAY,
+                                 ST_NODISPLAY, ST_NODISPLAY, ST_NODISPLAY, ST_NODISPLAY
+                                };  // current status of misc. subsystems
+
+static const char *status_chars[STATUS_MAX] = {
+  // group WiFi and transmission to internet servers
+  ".W0wA", // ST_WIFI_OFF, ST_WIFI_CONNECTED, ST_WIFI_ERROR, ST_WIFI_CONNECTING, ST_WIFI_AP
+  ".s1S",  // ST_SCOMM_OFF, ST_SCOMM_IDLE, ST_SCOMM_ERROR, ST_SCOMM_SENDING
+  ".m2M",  // ST_MADAVI_OFF, ST_MADAVI_IDLE, ST_MADAVI_ERROR, ST_MADAVI_SENDING
+  // group TTN (LoRa WAN)
+  ".t3T",  // ST_TTN_OFF, ST_TTN_IDLE, ST_TTN_ERROR, ST_TTN_SENDING
+  // group BlueTooth
+  ".B4b",  // ST_BT_OFF, ST_BT_CONNECTED, ST_BT_ERROR, ST_BT_CONNECTING
+  // group other
+  ".",     // ST_NODISPLAY
+  ".",     // ST_NODISPLAY
+  ".H7",   // ST_NODISPLAY, ST_HV_OK, ST_HV_ERROR
+};
+
+void set_status(int index, int value) {
+  if ((index >= 0) && (index < STATUS_MAX))
+    status[index] = value;
+  else
+    log(ERROR, "invalid parameters: set_status(%d, %d)", index, value);
+}
+
+char get_status_char(int index) {
+  if ((index >= 0) && (index < STATUS_MAX)) {
+    int idx = status[index];
+    if (idx < strlen(status_chars[index]))
+      return status_chars[index][idx];
+    else
+      log(ERROR, "string status_chars[%d] is too short, no char at index %d", index, idx);
+  } else
+    log(ERROR, "invalid parameters: get_status_char(%d)", index);
+  return '?';  // some error happened
+}
+
+void displayStatus(void) {
+  char output[17];  // max. 16 chars wide display + \0 terminator
+  const char *format = isLoraBoard ? "%c%c%c%c%c%c%c%c" : "%c %c %c %c %c %c %c %c";  // 8 or 16 chars wide
+  snprintf(output, 17, format,
+           get_status_char(0), get_status_char(1), get_status_char(2), get_status_char(3),
+           get_status_char(4), get_status_char(5), get_status_char(6), get_status_char(7)
+          );
+  displayStatusLine(output);
+}
+
 char *nullFill(int n, int digits) {
   static char erg[9];  // max. 8 digits possible!
   if (digits > 8) {
@@ -90,7 +138,7 @@ char *nullFill(int n, int digits) {
   return erg;
 }
 
-void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool connected, bool ble_active) {
+void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display) {
   if (!use_display) {
     if (!displayIsClear) {
       pu8x8->clear();
@@ -107,12 +155,7 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool conne
     char output[80];
     int TimeMin = TimeSec / 60;         // calculate number of minutes
     TimeMin = TimeMin % 1000;  // limit minutes to max. 999, roll over
-
-    // print the upper line including time and measured radation
-    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
-    pu8x8->print(ble_active ? '\x4A' : ' '); // 0x4A corresponds to Bluetooth symbol in selected font.
     pu8x8->setFont(u8x8_font_7x14_1x2_f);
-
     if (TimeSec < 60)    {    // < 1 minute -> display in seconds
       sprintf(output, "%2ds", TimeSec);
       pu8x8->print(output);
@@ -120,14 +163,11 @@ void DisplayGMC(int TimeSec, int RadNSvph, int CPM, bool use_display, bool conne
       sprintf(output, "%3d", TimeMin);
       pu8x8->print(output);
     }
-    sprintf(output, "%6d nSv/h", RadNSvph);
+    sprintf(output, "%7d nSv/h", RadNSvph);
     pu8x8->print(output);
     pu8x8->setFont(u8x8_font_inb33_3x6_n);
     pu8x8->drawString(0, 2, nullFill(CPM, 5));
   } else {
-    // print the upper line including BLE symbol in case of connection and measured radation
-    pu8x8->setFont(u8x8_font_open_iconic_embedded_1x1);
-    pu8x8->drawString(0, 2, ble_active ? "\x4A" : " "); // 0x4A corresponds to Bluetooth symbol in selected font.
     pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
     pu8x8->drawString(1, 2, nullFill(RadNSvph, 7));
     pu8x8->setFont(u8x8_font_px437wyse700b_2x2_f);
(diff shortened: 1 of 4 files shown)


multigeiger/display.cpp
```diff
@@ -81,6 +81,29 @@ void displayStatusLine(String txt) {
...
+
+const char *status_chars[STATUS_MAX] = {
+  " -wW0",  // ST_NODISPLAY, ST_WIFI_OFF, ST_WIFI_CONNECTING, ST_WIFI_CONNECTED, ST_WIFI_ERROR
+  " -bB1",  // ST_NODISPLAY, ST_BT_OFF, ST_BT_CONNECTING, ST_BT_CONNECTED, ST_BT_ERROR
+  " -lL2",  // ST_NODISPLAY, ST_LORA_OFF, ST_LORA_CONNECTING, ST_LORA_CONNECTED, ST_LORA_ERROR
+  " H3",    // ST_NODISPLAY, ST_HV_OK, ST_HV_ERROR
+  " -S4",   // ST_NODISPLAY, ST_SCOMM_IDLE, ST_SCOMM_SENDING, ST_SCOMM_ERROR
+  " -M5",   // ST_NODISPLAY, ST_MADAVI_IDLE, ST_MADAVI_SENDING, ST_MADAVI_ERROR
+  " -T6",   // ST_NODISPLAY, ST_TTN_IDLE, ST_TTN_SENDING, ST_TTN_ERROR
+  " -O7",   // ST_NODISPLAY, ST_TOILET_IDLE, ST_TOILET_SENDING, ST_TOILET_ERROR
```
ThomasWaldmann, line 94: this has to be seen, likely we will not need all these states / not all states make sense.

ideas:
- after boot, everything will display as "blank".
- lowercase == building something up (e.g. connecting)
- uppercase = buildup finished (e.g. connected)
- digit = error in that subsystem

multigeiger/display.cpp
```diff
@@ -81,6 +81,29 @@ void displayStatusLine(String txt) {
...
+};
+
+void displayStatus(void) {
+  char output[17];  // max. 16 chars wide display + \0 terminator
+  const char *format = isLoraBoard ? "%c%c%c%c%c%c%c%c" : "%c %c %c %c %c %c %c %c";  // 8 or 16 chars wide
```
ThomasWaldmann, line 99: that way, 8 status displays will fit on both displays.

multigeiger/display.h
```diff
@@ -4,8 +4,60 @@
...
+#define STATUS_TTN 6
+#define ST_TTN_IDLE 1
+#define ST_TTN_SENDING 2
+#define ST_TTN_ERROR 3
+
+#define STATUS_TOILET 7
+#define ST_TOILET_IDLE 1
+#define ST_TOILET_SENDING 2
+#define ST_TOILET_ERROR 3
+
```
ThomasWaldmann, line 56: please check which make sense and which don't.

rexfue:
Sounds good. LoRa and TTN are the same, aren't they? And I would rename TOILET to ADD_SERVER (additional server) or something similar. 
I think we should only introduce ONE additional server. Depending on the URL, this can either be the address of an MQTT server, the TOILET server or any other private server.

ThomasWaldmann:
@rexfue I agree, but this PR is based on the current state of the code and now we have "toilet" - changing this to an additional arbitrary server is out of scope of this PR.

Also, I removed the TOILET stuff again anyway because it is only active when some special #define is made, so only for debugging.

So we have 1 status display place now free for something in the future.

ThomasWaldmann:
@t-pi ^^^

sorry, had to remove your pretty BT indicator for now for consistency.

but we can make it more pretty again in future.

ThomasWaldmann:
@rexfue guess lora and ttn are not same.

one is the physical layer, the other a specific network (in other regions there might be different lora-based networks I guess?).

similar distinction is the physical wifi status (connected or not) and the sending to some specific server on the internet.

multigeiger/display.cpp
t-pi: Proposition: Have the same amount of letters per status display, with same status (e.g. uppercase) on same position. Fill unused positions with e.g. dot ("."). That way you can merge status changes more easily.

multigeiger/display.h
t-pi: s. above, maybe the status constants can be merged that way.

multigeiger/multigeiger.ino
```diff
@@ -265,6 +261,12 @@ void loop() {
   loop_duration = millis() - current_ms;
   iotWebConf.delay((loop_duration < LOOP_DURATION) ? (LOOP_DURATION - loop_duration) : 0);
 
+  static bool wifi_connected = false;
   wifi_connected = (iotWebConf.getState() == IOTWEBCONF_STATE_ONLINE);
+  status[STATUS_WIFI] = wifi_connected ? ST_WIFI_CONNECTED : ST_WIFI_CONNECTING;
+
+  status[STATUS_BT] = ST_NODISPLAY;  // TODO
+  status[STATUS_LORA] = ST_NODISPLAY;  // TODO
+  status[STATUS_HV] = ST_NODISPLAY;  // TODO
```
t-pi, line 270: Hmm, can't we have a more encapsulated way, like e.g. a function in display which is exposed, allowing to switch the status? I'm not too keen on all the includes and modifying variables in other modules... Would also be easier with fixed status numbers

multigeiger/display.cpp
ThomasWaldmann: Maybe we can not make this completely conistent as these are quite different subsystems.

But I can change NODISPLAY from space to dot.

multigeiger/transmission.cpp
```diff
@@ -220,45 +220,54 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
 
   if(isLoraBoard && sendToLora && (strcmp(appeui, "") != 0)) {    // send only, if we have LoRa credentials
     log(INFO, "Sending to TTN ...");
-    displayStatusLine("TTN");
+    status[STATUS_TTN] = ST_TTN_SENDING;
+    displayStatus();
     send_ttn_geiger(tube_nbr, dt, gm_counts);
     if (have_thp)
       send_ttn_thp(temperature, humidity, pressure);
+    status[STATUS_TTN] = ST_TTN_IDLE;
```
t-pi, line 264: same comment as above...

multigeiger/display.h
ThomasWaldmann: maybe they even need to be less consistent than now.

e.g. i am not sure whether lora really has a connecting / connected state or whether it is rather like udp "just send the stuff".

multigeiger/multigeiger.ino
ThomasWaldmann: could be done, then status would not need to be global any more.

multigeiger/display.cpp
t-pi: What I meant was to make the position of the states consistent, e.g.
// NODISPLAY, OFF, IDLE, CONNECTING, CONNECTED, ERROR
  " -.wW0",  // ST_NODISPLAY, ST_WIFI_OFF, -, ST_WIFI_CONNECTING, ST_WIFI_CONNECTED, ST_WIFI_ERROR
  " -b.B1",  // ST_NODISPLAY, ST_BT_OFF, ST_BT_IDLE, -, ST_BT_CONNECTED, ST_BT_ERROR
  " -.lL2",  // ST_NODISPLAY, ST_LORA_OFF, -, ST_LORA_CONNECTING, ST_LORA_CONNECTED, ST_LORA_ERROR
  " ...H3",    // ST_NODISPLAY, -, -, -, ST_HV_OK, ST_HV_ERROR
  " -s.S4",   // ST_NODISPLAY, ST_SCOMM_OFF, ST_SCOMM_IDLE, -, ST_SCOMM_SENDING, ST_SCOMM_ERROR
...

That way, further functions, like e.g. inverting an error, could be implemented more consistently

t-pi: One addition, maybe WIFI could indicate the Access Point active with "a" instead of idle...

multigeiger/multigeiger.ino
ThomasWaldmann: done, will update PR soon.

multigeiger/display.cpp
ThomasWaldmann: fixed, now always in this order: off 0, ok 1, error 2.

off is same as nodisplay now (and displayed as `.`).

ok is uppercase now, error is numeric.

> 2 is custom, depending on the subsystem.
lowercase is used for connecting.

the network states are a bit different: here lowercase is for idle and enabled, uppercase is for sending.

ThomasWaldmann:
Current status line outputs (this is what already works):

```
w . . H . . . .  # wifi connecting, HV ok
W . . H . . . .  # wifi connected, HV ok
. . . 4 . . . .  # error 4 == HV bad
W . . H c m . .  # wifi connected, s.comm. madavi enabled, but idle
W . . H C m . .  # wifi connected, s.comm sending. madavi idle
```

LoRA, BT, TTN is still todo.

Also WiFi "AP" status.

ThomasWaldmann:
Hmm, I guess i shift HV status to the right. It now sits in between network stuff.

ThomasWaldmann:
guess this could be collapsed / merged now.

it will need more work for BT and LORA, but that can be done in separate PRs (BT is not merged yet and I can't test LORA myself).

@rexfue @t-pi do you agree or is there still todo for me?

rexfue:
It's OK for me, merge.

multigeiger/display.cpp
ThomasWaldmann: AP status now implemented with "A" in place of "W".

ThomasWaldmann:
@t-pi ok, please fix that in your BLE PR.

about grouping: yeah, I also thought about that, somehow a matter of taste I guess.

i grouped together:
- physical stuff (WiFi, BT, LoRa)
- network/transmission stuff (SCOMM, MADAVI, TTN)
- other stuff (unused, HV)

@t-pi suggest to group like:
- wifi/network based stuff (WiFi, SCOMM, MADAVI)
- bt based stuff (BT, ...?)
- lora/lora network based stuff (LoRa, TTN)

@rexfue what do you prefer?

t-pi:
Okay, I see your logic. Also fine with me.
However I guess most people would group TTN more with Lora, as I did initially.
Up to you, @rexfue :)

ThomasWaldmann:
If someone would use multigeiger in the US or China, would they also use TTN?

rexfue:
I think so, its world wide

t-pi:
Maybe we could the Lora status so far just for "isLoraBoard" or not? Doesn't make much sense on the display, though, only for an overall querying of status e.g. for remote display over BLE.

ThomasWaldmann:
had a look at their map. yes, seems pretty international.

so, I'll drop `LORA` status and only have a `TTN` status.
and if the board does not have LoRa hardware, that would be ST_TTN_OFF / ST_NODISPLAY.

rexfue:
ok, that sounds good

ThomasWaldmann:
@t-pi i now grouped as to your suggestion - due to the removal of `LoRa` my grouping felt strange...

also fixed the ST_BT_CONNECTABLE.

t-pi:
👍 thanks!

ThomasWaldmann:
collapsed everything together, will merge after tests have completed.

rexfue:
wouldn't it be better to start error numbers at 1? Because error number 0 doesn't sound like an error.

ThomasWaldmann:
error number? do you mean the ST_*_* codes? `0` there means `OFF / NODISPLAY` which is a reasonable initial state.

rexfue:
No, the status bar error numbers. You have WiFi = 0, SCOMM = 1 ...  until H = 7. Shouldn't it be shifted to 1 ... 8?

ThomasWaldmann:
The idea behind using numbers (which correspond directly to the index in our status_chars array) is that one can easily tell if there is a problem. If you see numbers, you have a problem.

ThomasWaldmann:
... and that is 0-based, as usual in C (or other programming languages).

We could offset everything there to be 1-based, but I currently don't see a pressing need.

PR #269, opened 2020-05-11
check for http transmission errors, update status, fixes #265

### LINKED ISSUE #265: status display: display http transmission errors
the status display has ST_*_ERROR status values for SCOMM and MADWIFI.

if the transmission code detects an unsuccessful transmission (http result is not `200 OK`), it should call `set_status()` accordingly.

ThomasWaldmann:
kind of trivial, but not practically tested yet.


ThomasWaldmann:
merged this now to avoid conflicts.

PR #270, opened 2020-05-11
rearranged wifi_status and block

rexfue:
sendig, if not connected

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -200,7 +200,10 @@ void read_THP(unsigned long current_ms,
 }
 
 void transmit(unsigned long current_ms, unsigned long current_counts, unsigned long gm_count_timestamp, unsigned long current_hv_pulses,
-              bool have_thp, float temperature, float humidity, float pressure) {
+              bool have_thp, float temperature, float humidity, float pressure, bool wificonnected) {
+  if( !wificonnected) {
+    return;
+  }
   static unsigned long last_counts = 0;
   static unsigned long last_hv_pulses = 0;
   static unsigned long last_timestamp = millis();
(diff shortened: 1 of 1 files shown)


multigeiger/multigeiger.ino
```diff
@@ -274,16 +277,16 @@ void loop() {
   if (Serial_Print_Mode == Serial_Statistics_Log)
     statistics_log(gm_counts, gm_count_time_between);
 
-  transmit(current_ms, gm_counts, gm_count_timestamp, hv_pulses, have_thp, temperature, humidity, pressure);
+  int wifi_status = update_wifi_status();
+
+  transmit(current_ms, gm_counts, gm_count_timestamp, hv_pulses, have_thp, temperature, humidity, pressure, (wifi_status == ST_WIFI_CONNECTED));
```
ThomasWaldmann, line 282: I'ld rather put the `if` here. See also my comments in the ticket.

multigeiger/multigeiger.ino
```diff
@@ -200,7 +200,10 @@ void read_THP(unsigned long current_ms,
 }
 
 void transmit(unsigned long current_ms, unsigned long current_counts, unsigned long gm_count_timestamp, unsigned long current_hv_pulses,
-              bool have_thp, float temperature, float humidity, float pressure) {
+              bool have_thp, float temperature, float humidity, float pressure, bool wificonnected) {
+  if( !wificonnected) {
+    return;
```
ThomasWaldmann, line 205: guess this kills a lora device that currently has no wifi connection (which is kind of expected).

multigeiger/multigeiger.ino
ThomasWaldmann: hmm, no, it's more difficult.

transmit needs (usually) either lora or wifi, in theory it could even use wifi and lora.

multigeiger/multigeiger.ino
ThomasWaldmann: so there should be a "if wificonnected" around the transmission stuff needing wifi and another "if lora available" around the transmission stuff needing lora.

rexfue: if wificonnected -> OK
But we don't have a 'if lora available'. If we have lora hardware, we have lora available. If we cannot join, we get an error but we have to try the join in any way.
So if 'sendToLora' is true, we always (try to) send.

ThomasWaldmann:
superseded by #297.

PR #271, opened 2020-05-11
fix transmission to http: URLs, fixes #262

### LINKED ISSUE #262: TLS support doesn't work with http:// url
Changed the urls for madvi  and sensor.community from https:// to http://. Now there comes an error response:
```
[E][ssl_client.cpp:87] start_ssl_client(): Connect to Server failed!
[E][WiFiClientSecure.cpp:132] connect(): start_ssl_client: -1
```
If I use https, all works fine.

ThomasWaldmann:
this is practically tested using a http: as well as a https: toilet URL, it worked for both.

### DIFF
--- a/multigeiger/transmission.cpp
+++ b/multigeiger/transmission.cpp
@@ -61,7 +61,10 @@ void setup_transmission(const char *version, char *ssid, bool loraHardware) {
 }
 
 void prepare_http(HttpsClient *client, const char *host) {
-  client->hc->begin(*client->wc, host);
+  if (host[4] == 's')  // https
+    client->hc->begin(*client->wc, host);
+  else  // http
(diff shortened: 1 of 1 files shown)


rexfue:
Super. It works. Thanks. And it's definite faster!
Please make http:// as default because of the restrictions at sensor.community.

ThomasWaldmann:
OK, I'll change it to http for now.

BUT:
- if they do not want to offer https, why do they offer https? seems inconsistent.
- also, i don't think https is really an issue, considering how few multigeigers there are.
- it should be tried to optimize the server maybe. https is kind of state of the art.

ThomasWaldmann:
what about madavi? also change to http or keep https?

PR #272, opened 2020-05-11
transmission: http, toilet, custom server

ThomasWaldmann:
- use http for now, except for testing
- generalized custom server
- added comments about toilet usage :-)

### DIFF
--- a/multigeiger/transmission.cpp
+++ b/multigeiger/transmission.cpp
@@ -24,8 +24,12 @@
 // use http for now, server operator tells there are performance issues with https.
 #define SENSORCOMMUNITY "http://api.sensor.community/v1/push-sensor-data/"
 
-// Note: Custom TOILET URLs from https://ptsv2.com/ work with https and http.
-#define TOILET "https://ptsv2.com/t/rk9pr-1582220446/post"
+// Send http(s) post requests to a custom server
+// Note: Custom toilet URLs from https://ptsv2.com/ can be used for debugging
+// and work with https and http.
+#define CUSTOMSRV "https://ptsv2.com/t/xxxxx-yyyyyyyyyy/post"
+// Get your own toilet URL and put it here before setting this to true.
+#define SEND2CUSTOMSRV false
 
 static String http_software_version;
 static unsigned int lora_software_version;
@@ -37,7 +41,7 @@ typedef struct https_client {
   HTTPClient *hc;
 } HttpsClient;
 
-static HttpsClient c_madavi, c_sensorc, c_toilet;
+static HttpsClient c_madavi, c_sensorc, c_customsrv;
 
 void setup_transmission(const char *version, char *ssid, bool loraHardware) {
   chipID = String(ssid);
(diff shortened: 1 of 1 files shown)


multigeiger/transmission.cpp
```diff
@@ -17,9 +17,19 @@
 #include "ca_certs.h"
 
 // Hosts for data delivery
-#define MADAVI "https://api-rrd.madavi.de/data.php"
-#define SENSORCOMMUNITY "https://api.sensor.community/v1/push-sensor-data/"
-#define TOILET "https://ptsv2.com/t/rk9pr-1582220446/post"
+
+// use http for now, server operator tells there are performance issues with https.
+#define MADAVI "http://api-rrd.madavi.de/data.php"
```
ThomasWaldmann, line 22: true also?

multigeiger/transmission.cpp
```diff
@@ -17,9 +17,19 @@
...
+
+// use http for now, server operator tells there are performance issues with https.
+#define SENSORCOMMUNITY "http://api.sensor.community/v1/push-sensor-data/"
```
ThomasWaldmann, line 25: true. sadly.

multigeiger/transmission.cpp
```diff
@@ -224,13 +234,13 @@ void transmit_data(String tube_type, int tube_nbr, unsigned int dt, unsigned int
...
-  rc2 = have_thp ? send_http_thp(&c_toilet, TOILET, temperature, humidity, pressure, XPIN_NO_XPIN) : 200;
-  toilet_ok = (rc1 == 200) && (rc2 == 200);
-  log(INFO, "SENT TO TOILET, status: %s, http: %d %d", toilet_ok ? "ok" : "error", rc1, rc2);
+  #if SEND2CUSTOMSRV
+  bool customsrv_ok;
+  log(INFO, "Sending to CUSTOMSRV ...");
+  rc1 = send_http_geiger(&c_customsrv, CUSTOMSRV, dt, hv_pulses, gm_counts, cpm, XPIN_NO_XPIN);
+  rc2 = have_thp ? send_http_thp(&c_customsrv, CUSTOMSRV, temperature, humidity, pressure, XPIN_NO_XPIN) : 200;
+  customsrv_ok = (rc1 == 200) && (rc2 == 200);
+  log(INFO, "Sent to CUSTOMSRV, status: %s, http: %d %d", customsrv_ok ? "ok" : "error", rc1, rc2);
```
ThomasWaldmann, line 243: Less smelly. :-)

multigeiger/transmission.cpp
```diff
@@ -17,9 +17,19 @@
...
+
+// use http for now, server operator tells there are performance issues with https.
+#define SENSORCOMMUNITY "http://api.sensor.community/v1/push-sensor-data/"
+
+// Send http(s) post requests to a custom server
+// Note: Custom toilet URLs from https://ptsv2.com/ can be used for debugging
+// and work with https and http.
+#define CUSTOMSRV "https://ptsv2.com/t/xxxxx-yyyyyyyyyy/post"
+// Get your own toilet URL and put it here before setting this to true.
+#define SEND2CUSTOMSRV false
```
ThomasWaldmann, line 32: ok to define SEND2CUSTOMSRV here?

multigeiger/transmission.cpp
rexfue: :-) :-)

multigeiger/transmission.cpp
rexfue: OK, that's fine.

multigeiger/transmission.cpp
ThomasWaldmann: @rexfue do you know? is this the same server or admin and does the same issue apply there?

rexfue: Sorry, I don't know for sure. But I guess, there are the same restrictions.

ThomasWaldmann: It's a different IP, thus a different server.

I'll remove the comment. Guessing is not helpful here.

Issue #227, opened 2020-03-09
learning git / github workflow


ThomasWaldmann:
this is about the workflow we also use for multigeiger:
https://www.youtube.com/watch?v=8UguQzmswC4

there is also some stuff in our docs:
https://multigeiger.readthedocs.io/en/latest/development.html

ThomasWaldmann:
@WwwBoehriDe @rexfue @ecocurious ^^^

ThomasWaldmann:
guess everybody interested has the links now.

t-pi:
Dunno, might be good to keep this open for new contributors...?
I found e.g. this YouTube video particularily enlightening:
https://www.youtube.com/watch?v=M7ZYkjOWr6g

ThomasWaldmann:
Make a PR to add it to contributor docs? See `docs/...`

t-pi:
ok, will have a look, thanks

PR #274, opened 2020-05-11
add bme680 library for basic THP readout of BME680

t-pi:
After wondering some time why my THP stopped working after the FW update, I noticed it's an BME680 board. This PR adds BME680 lib to allow transparent readout of THP only from BME680 as well as from BME280.
Future option to read and log gas resistance value for air quality changes not considered here.

### DIFF
--- a/multigeiger/thp_sensor.cpp
+++ b/multigeiger/thp_sensor.cpp
@@ -4,29 +4,69 @@
 
 #include <Adafruit_Sensor.h>
 #include <Adafruit_BME280.h>
+#include <Adafruit_BME680.h>
 
 #include "log.h"
 #include "thp_sensor.h"
 
-static bool have_thp = false;
+static int type_thp = 0;
 
-Adafruit_BME280 bme;
+Adafruit_BME280 bme280;
+Adafruit_BME680 bme680;
 
 bool setup_thp_sensor(void) {
-  have_thp = (bool) bme.begin(BME280_ADDRESS);
-  if (!have_thp)
-    have_thp = (bool) bme.begin(BME280_ADDRESS_ALTERNATE);
-  if (have_thp)
-    log(INFO, "BME_Status: %d  ID:%0X", (int) have_thp, bme.sensorID());
-  else
-    log(INFO, "BME_Status: %d  ID:None", (int) have_thp);
-  return have_thp;
+  // BME280
+  if (bme280.begin(BME280_ADDRESS))
+    type_thp = 280;
+  else if (bme280.begin(BME280_ADDRESS_ALTERNATE))
+    type_thp = 280;
+
+  // BME680
+  if (type_thp == 0) {
+    if (bme680.begin(BME680_I2C_ADDR_PRIMARY))
+      type_thp = 680;
+    else if (bme680.begin(BME680_I2C_ADDR_SECONDARY))
+      type_thp = 680;
+  }
+
+  switch (type_thp) {
+  case 680:
+    // Set up oversampling and filter initialization
+    bme680.setTemperatureOversampling(BME680_OS_8X);
+    bme680.setHumidityOversampling(BME680_OS_2X);
+    bme680.setPressureOversampling(BME680_OS_4X);
+    bme680.setIIRFilterSize(BME680_FILTER_SIZE_3);
+    bme680.setGasHeater(300, 150); // 300*C for 150 ms
+    log(INFO, "BME_Status: ok,  ID: BME680");
+    break;
+  case 280:
+    log(INFO, "BME_Status: ok,  ID: BME280");
+    break;
+  default:
+    log(INFO, "BME_Status: not found");
+    break;
+  }
+  return (type_thp > 0);
 }
 
 bool read_thp_sensor(float *temperature, float *humidity, float *pressure) {
-  *temperature = have_thp ? bme.readTemperature() : 0.0;
-  *humidity = have_thp ? bme.readHumidity() : 0.0;
-  *pressure = have_thp ? bme.readPressure() : 0.0;
-  return have_thp;
+  if (type_thp == 280) {
+    *temperature = bme280.readTemperature();
+    *humidity = bme280.readHumidity();
+    *pressure = bme280.readPressure();
+  } else if (type_thp == 680) {
+    if (!bme680.performReading()) {
+      log(INFO, "BME680: Failed to perform reading");
+      return false;
+    }
+    *temperature = bme680.temperature;
+    *humidity = bme680.humidity;
+    *pressure = bme680.pressure;
+  } else {
+    *temperature = 0.0;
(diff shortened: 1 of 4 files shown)


t-pi:
Ah, and I changed THP log from DEBUG to INFO, because the missing logs took me also a while to figure out correct functionality...

multigeiger/thp_sensor.cpp
```diff
@@ -4,29 +4,61 @@
...
+    *pressure = bme280.readPressure();
+  } else if (type_thp == 6) {
+    if (!bme680.performReading()) {
+      log(INFO, "BME680: Failed to perform reading");
+      return false;
+    }
+    *temperature = bme680.temperature;
+    *humidity = bme680.humidity;
+    *pressure = bme680.pressure;
+  }
```
ThomasWaldmann, line 61: maybe move the 0.0 values into the else-branch for the case we do not have any sensor and call read_the_sensor.

then at least the values get set to **some** (obviously wrong) value instead of maybe not being initialized.

multigeiger/thp_sensor.cpp
```diff
@@ -4,29 +4,61 @@
...
 #include <Adafruit_Sensor.h>
 #include <Adafruit_BME280.h>
+#include <Adafruit_BME680.h>
 
 #include "log.h"
 #include "thp_sensor.h"
 
-static bool have_thp = false;
+static bool have_thp;
+static int type_thp = 0;
```
ThomasWaldmann, line 13: maybe one could do it with just one of them. 0 == no sensor, 280 = bme280, 680 == bme680.

multigeiger/thp_sensor.cpp
```diff
@@ -4,29 +4,61 @@
...
 
-Adafruit_BME280 bme;
+Adafruit_BME280 bme280;
+Adafruit_BME680 bme680;
 
 bool setup_thp_sensor(void) {
```
ThomasWaldmann, line 18: this could return the type of the detected sensor or 0 for none.

multigeiger/thp_sensor.cpp
```diff
@@ -4,29 +4,61 @@
...
+  have_thp = (bool) bme680.begin(BME680_I2C_ADDR_PRIMARY);
+  if (!have_thp)
+    have_thp = (bool) bme680.begin(BME680_I2C_ADDR_SECONDARY);
+  if (have_thp) {
+    // Set up oversampling and filter initialization
+    bme680.setTemperatureOversampling(BME680_OS_8X);
+    bme680.setHumidityOversampling(BME680_OS_2X);
+    bme680.setPressureOversampling(BME680_OS_4X);
+    bme680.setIIRFilterSize(BME680_FILTER_SIZE_3);
+    bme680.setGasHeater(300, 150); // 300*C for 150 ms
```
ThomasWaldmann, line 38: is this needed even if you do not use the gas sensor yet?

platformio-example.ini
```diff
@@ -25,6 +25,7 @@ framework = arduino
 monitor_speed=115200
 lib_deps=
   U8g2
+  Adafruit BME680 Library
```
rexfue, line 28: Please also include the library in the Readme

ThomasWaldmann: maybe the readme could just say "please install all libraries you see in platformio-example.ini lib_deps"?

rexfue: Ah, that's a good idea.   ".. and always use the newest versions."

t-pi: Will do

multigeiger/thp_sensor.cpp
t-pi: I prefer to have the sensor fully initialized. For the portable version, an air quality index might be interesting, too...

multigeiger/thp_sensor.cpp
t-pi: Okay, will simplify

multigeiger/thp_sensor.cpp
t-pi: will do.

multigeiger/thp_sensor.cpp
t-pi: Will do, too.

t-pi:
Could anyone check the sequential OR initialisation with BME280, please? 
I only have BME680 here. Thanks.

ThomasWaldmann:
there are some unrelated commits in your PR.

t-pi:
Unrelated commits are probably from my git babysteps...
I do an 
```
git checkout master
git pull --rebase upstream master
```
to rebase to upstream master. 
Ah! And I pushed to my fork master. Maybe that's why all intermediate commits are shown above...?

t-pi:
And one moment please, found an issue. Fixing...

ThomasWaldmann:
I usually keep my master branch up-to-date and in sync to this repo.
For own work, I create a feature-branch (as you did).
If I want to rebase, I just do (having my feature branch checked out):
```
git rebase -i master
```

ThomasWaldmann:
Before doing more work, please fix the issue with the unrelated changesets - one can not review your overall changes right now.

t-pi:
I am following the aforementioned procedure from [here](https://www.youtube.com/watch?v=M7ZYkjOWr6g). He is linking the upstream to my local checkout. Last time it worked better, so I guess this intermediate push to my fork inserted the unrelated commits...

How do I get rid of them, any idea?

ThomasWaldmann:
Guess I'ld first make a local copy (using `git clone repo repo2`) to avoid making it worse.
The I'ld try what I have already described above and check whether that cleans it up.
Likely you need to `git push -f ...` after the rebase.

ThomasWaldmann:
And the usual workflow was already discussed in #227.

t-pi:
Phew, looks good to me...? 🤞

rexfue:
Checked - identifies BME280 and reads values correctly.

rexfue:
There is only a little problem: sensor.comunity doesn't support BME680 at the moment !
If we send only temp, humidity and press we can send 'incognito' as BME280.

ThomasWaldmann:
When sending "final" values, like t/h/p, I guess the sensor is not relevant.

It's a different situation when sending CPM values of a GM tube. Without knowing the tube, the CPM value would be useless.

rexfue:
Yes we need the type:  sensor.comm sends the sensor type with the current data.  Also madavi needs the sensor name, which is in the file name of the archive. 
I'll ask Rajko if he can add it.

t-pi:
okay, @rexfue, thanks. Currently I did not change the designation of my sensor in devices.sensor.community, so I am indeed sending "incognito". I know Rajko isn't keen on adding the BME680's gas sensing capability, as it's values depend on the sensor's history. I guess for the THP it should not matter, though, as Thomas already mentioned.

PR #284, opened 2020-05-13
docs: add usage chapter, description of OLED display, fixes #264

### LINKED ISSUE #264: document status display
#257 implements a status display.

after that, there will be some more work on it relating to LoRa status and BT status (after bluetooth work is merged).

As showing a bunch of single letters/digits is not self-explaining, we should have some .rst docs about it.

ThomasWaldmann:
i made this using the bigger display on the wifi-only board.

@rexfue can you check whether the description is ok for the wifi/lora board also?

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -15,6 +15,7 @@ MultiGeiger Documentation
    :maxdepth: 2
 
    changes
+   usage
    faq
    resources
    development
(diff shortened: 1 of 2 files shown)


rexfue:
OK, the description almost fits - except the top line left: the time is NOT shown on the small display

PR #285, opened 2020-05-13
status display: keep showing transmission errors

ThomasWaldmann:
in case of some unsuccessful transmission attempt, the error code was
only shown for a short time and then replaced by idle status again.

fixed, now the error code is shown until the transmission is retried.

also removed some unneeded status display related code for BT and LoRa,
the initial state is "NODISPLAY" anyway.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -42,21 +42,21 @@ Positions:
 - 1: sensors.community transmission
 
   - ``.``: off (not configured / enabled)
-  - ``s``: idle
   - ``S``: sending
-  - ``1``: transmission failed
+  - ``s``: idle (shown after successful sending)
+  - ``1``: sending failed (shown after trying to send)
 - 2: madavi transmission
 
   - ``.``: off (not configured / enabled)
-  - ``m``: idle
   - ``M``: sending
-  - ``2``: transmission failed
+  - ``m``: idle (shown after successful sending)
+  - ``2``: sending failed (shown after trying to send)
 - 3: TTN ("The Things Network")
 
-  - ``.``: off (no LoRa hardware, not configured, not enabled)
-  - ``t``: idle
+  - ``.``: off (not configured, not enabled, no LoRa hardware)
   - ``T``: sending
-  - ``3``: some error happened
+  - ``t``: idle (shown after successful sending)
+  - ``3``: sending failed (shown after trying to send)
 - 4: reserved for BT
 - 5: unused
 - 6: unused
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -252,6 +252,9 @@ void loop() {
   read_THP(current_ms, &have_thp, &temperature, &humidity, &pressure);
 
   read_hv(&hv_error, &hv_pulses);
+  set_status(STATUS_HV, hv_error ? ST_HV_ERROR : ST_HV_OK);
+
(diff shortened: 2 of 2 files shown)


docs/source/usage.rst
```diff
@@ -42,21 +42,21 @@ Positions:
...
+  - ``2``: sending failed (shown after trying to send)
 - 3: TTN ("The Things Network")
 
-  - ``.``: off (no LoRa hardware, not configured, not enabled)
-  - ``t``: idle
+  - ``.``: off (not configured, not enabled, no LoRa hardware)
   - ``T``: sending
-  - ``3``: some error happened
+  - ``t``: idle (shown after successful sending)
+  - ``3``: sending failed (shown after trying to send)
```
ThomasWaldmann, line 59: this needs to be implemented by somebody with LoRa hardware.

multigeiger/multigeiger.ino
```diff
@@ -252,6 +252,9 @@ void loop() {
   read_THP(current_ms, &have_thp, &temperature, &humidity, &pressure);
 
   read_hv(&hv_error, &hv_pulses);
+  set_status(STATUS_HV, hv_error ? ST_HV_ERROR : ST_HV_OK);
+
+  int wifi_status = update_wifi_status();
 
   display(current_ms, gm_counts, gm_count_timestamp, hv_pulses);
```
ThomasWaldmann, line 259: i moved status update to 255 / 257, so it gets immediately displayed by 259.

multigeiger/multigeiger.ino
```diff
@@ -266,15 +269,4 @@ void loop() {
   long loop_duration;
   loop_duration = millis() - current_ms;
   iotWebConf.delay((loop_duration < LOOP_DURATION) ? (LOOP_DURATION - loop_duration) : 0);
-
-  int wifi_status = update_wifi_status();
-
-  set_status(STATUS_SCOMM, ((wifi_status == ST_WIFI_CONNECTED) && sendToCommunity) ? ST_SCOMM_IDLE : ST_SCOMM_OFF);
-  set_status(STATUS_MADAVI, ((wifi_status == ST_WIFI_CONNECTED) && sendToMadavi) ? ST_MADAVI_IDLE : ST_MADAVI_OFF);
```
ThomasWaldmann, line 273: this is not needed here and even harmful. we just want to keep whatever the transmission code has set as status.

multigeiger/multigeiger.ino
```diff
@@ -266,15 +269,4 @@ void loop() {
...
-
-  set_status(STATUS_TTN, sendToLora ? ST_TTN_IDLE : ST_TTN_OFF);
```
ThomasWaldmann, line 275: "OFF" is default anyway. And, see previous comment, updating the status needs to be done AFTER trying to send.

After successful sending -> IDLE status.

After erroneous sending attempt -> ERROR status.

multigeiger/multigeiger.ino
```diff
@@ -266,15 +269,4 @@ void loop() {
...
-
-  set_status(STATUS_BT, ST_NODISPLAY);  // TODO
```
ThomasWaldmann, line 277: This needs to be solved when BT is added. Default status is "OFF", so no problem for now.

ThomasWaldmann:
I quickly merged that to not get into my own way with more incoming changes.

@rexfue please have a look into the comments here, might be useful for your future lora status work.

PR #296, opened 2020-05-14
show lora error on status display

### LINKED ISSUE #266: status display: display TTN transmission errors?
the status display for TTN has a ST_TTN_ERROR status.

can we detect an unsuccessful TTN transmission?

if so, we should call `set_status()` accordingly.

if not, we should remove the ST_TTN_ERROR status and indicator.

rexfue:
fixes #266


ThomasWaldmann:
LGTM :)

PR #300, opened 2020-05-14
display cleanup

### LINKED ISSUE #290: cleanup DisplayGMC for better consistency
DisplayGMC has very different code for the bigger and the smaller display.

One is using sprintf and print, the other is using nullFill and drawString.

I'ld suggest:
- using sprintf for both
- removing nullFill (which just uses sprintf internally)
- decide on either using drawString or print

### LINKED ISSUE #298: does status output overwrite "cpm" text?
for the lora wifi stick, both output to line 5...

note: i do not mean the cpm value (digits), but the "cpm" text (letters).

### LINKED ISSUE #299: extend time display
for wifi kit / bigger display we display the uptime of the multigeiger.

currently:
- 0..60s
- 0..999 (minutes), then roll around

proposed:
- 0..60s
- 0..99m
- 0..99h
- 0..99d, then roll around

ThomasWaldmann:
fixes #299 
fixes #290 
fixes #298


rexfue:
check on small display OK - merge.

PR #302, opened 2020-05-14
status display: add init status ("?"), docs, fixes #295

### LINKED ISSUE #295: Status display after reset
Should the characters for the servers (s, m, t) not appear directly after reset if enabled?

ThomasWaldmann:


rexfue:
Looks good. I would prefer exclamation mark instead of question mark. But that's not important.

t-pi:
Actually, ! would feel more like an error to me. I'd prefer ?

rexfue:
OK, 2 : 1  - ? wins :-)

ThomasWaldmann:
For the time when `?` is displayed, it is unclear whether it can actually make that connection.
It is enabled, but we did not try yet. So `?` is pretty right for that, IMHO.

And yes, I would also see `!` as a warning or error rather.

PR #306, opened 2020-05-15
fix flashing issues

ThomasWaldmann:


multigeiger/speaker.cpp
```diff
@@ -111,43 +115,60 @@ void IRAM_ATTR isr_audio() {
...
+    sequence[1] = 1;  // volume
+    sequence[2] = led_tick ? (high ? 1 : -1) : -1;  // LED
+    sequence[3] = 4;  // duration_ms
+    // "off"
+    sequence[4] = speaker_tick ? 0 : -1;
+    sequence[5] = 0;
+    sequence[6] = led_tick ? (high ? 0 : -1) : -1;
+    sequence[7] = 0;  // END
+  } else
+    sequence = NULL;
```
ThomasWaldmann, line 141: This is the important part. If the sequence is NULL, the ISR will be finished very quickly and not access any library code that resides in flash memory.

multigeiger/speaker.cpp
```diff
@@ -84,14 +87,15 @@ void IRAM_ATTR isr_audio() {
...
-  } else {
+  } else if (frequency_mHz == 0) {  // speaker off
     // frequency_mHz == 0 -> disable sound output
     // stop any PWM signals
     mcpwm_stop(MCPWM_UNIT_0, MCPWM_TIMER_0);
     // keep A high and B low (we have a piezo, no current flowing)
     mcpwm_set_signal_high(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_A);
     mcpwm_set_signal_low(MCPWM_UNIT_0, MCPWM_TIMER_0, MCPWM_OPR_B);
   }
+  // frequency_mHz == -1 -> don't touch pwm/speaker
```
ThomasWaldmann, line 98: slight optimization: if user switches off speaker ticks, it won't do any PWM calls here.

multigeiger/webconf.cpp
```diff
@@ -120,6 +121,12 @@ void handleRoot(void) {  // Handle web requests to "/" path.
     "</body>"
     "</html>\n";
   server.send(200, "text/html;charset=UTF-8", index);
+  // looks like user wants to do some configuration or maybe flash firmware.
+  // while accessing the flash, we need to turn ticking off to avoid exceptions.
+  // user needs to save the config (or flash firmware + reboot) to turn it on again.
+  // note: it didn't look like there is an easy way to put this call at the right place
+  // (start of fw flash / start of config save) - this is why it is here.
+  tick_enable(false);
```
ThomasWaldmann, line 129: always go to `/` URL first to turn off the ticking.

rexfue:
Good job, thanks.
FYI:  It only crashes if I use OTA, if I change one of the settings via web interface, it won't crash. Maybe, because i have the big (8MB) partition scheme.

ThomasWaldmann:
Yeah, either that or it is just a bit random / luck. I just needed to "save as is" and boom...

Issue #208, opened 2020-02-20
u8x8 clear screen bug?


ThomasWaldmann:
there is some extra code that clears some display lines although there was a `u8x8.clear()` call right before.

check whether there is a bug in u8x8 and if so, file it upstream.

rexfue:
I try to get the schematics of the board to see what type of display this is. Maybe we selected the wrong one.

ThomasWaldmann:
Looks like the extra lines are not needed, will get fixed by #312.

PR #312, opened 2020-05-17
small changes to display, displaystatus

t-pi:
pulled out from ble_connectivity PR, as out of scope there

### DIFF
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -28,25 +28,28 @@ void display_start_screen(void) {
   pu8x8->clear();
   if (isLoraBoard) {
     // Display is only 4 lines by 8 characters; lines counting from 2 to 5
-    pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);        // u8x8 can handle only 8x8, smaller font has no advantage
+    pu8x8->setFont(u8x8_font_victoriamedium8_r);
     for (int i = 2; i < 6; i++) {
       pu8x8->drawString(0, i, "        ");  // clear all 4 lines
     }
   }
 
   if (isLoraBoard) {
-    pu8x8->drawString(0, 2, "Geiger-");
-    pu8x8->drawString(0, 3, " Counter");
-    pu8x8->drawString(0, 4, "Version:");
+    pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
+    pu8x8->drawString(0, 2, "Multi-");
+    pu8x8->drawString(0, 3, " Geiger");
+    pu8x8->setFont(u8x8_font_victoriamedium8_r);
     snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
-    pu8x8->drawString(0, 5, line);
+    pu8x8->drawString(0, 4, line);
   } else {
     pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
-    pu8x8->drawString(0, 0, "Geiger-Counter");
-    pu8x8->drawString(0, 2, "==============");
+    pu8x8->drawString(0, 0, "Multi-Geiger-");
+    pu8x8->drawString(0, 1, " Counter");
+    pu8x8->setFont(u8x8_font_victoriamedium8_r);
+    pu8x8->drawString(0, 2, "Info:boehri.de");
+    pu8x8->drawString(0, 4, "Version:");
     snprintf(line, 15, "%s", VERSION_STR);  // 14 chars + \0 termination
-    pu8x8->drawString(0, 4, line);
-    pu8x8->drawString(0, 6, "Info:boehri.de");
+    pu8x8->drawString(0, 5, line);
   }
   displayIsClear = false;
 };
@@ -74,7 +77,7 @@ void displayStatusLine(String txt) {
   if (txt.length() == 0)
     return;
   int line = isLoraBoard ? 5 : 7;
-  pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
+  pu8x8->setFont(u8x8_font_victoriamedium8_r);
   clearDisplayLine(line);
   pu8x8->drawString(0, line, txt.c_str());
 }
@@ -91,7 +94,7 @@ static const char *status_chars[STATUS_MAX] = {
   // group TTN (LoRa WAN)
   ".t3T?",  // ST_TTN_OFF, ST_TTN_IDLE, ST_TTN_ERROR, ST_TTN_SENDING, ST_TTN_INIT
   // group BlueTooth
-  ".B4b?",  // ST_BT_OFF, ST_BT_CONNECTED, ST_BT_ERROR, ST_BT_CONNECTING, ST_BT_INIT
+  ".B4b?",  // ST_BT_OFF, ST_BT_CONNECTED, ST_BT_ERROR, ST_BT_CONNECTABLE, ST_BT_INIT
   // group other
   ".",      // ST_NODISPLAY
   ".",      // ST_NODISPLAY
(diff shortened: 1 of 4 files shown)


t-pi:
did also some fiddling with the fonts, as indeed the "font_amstrad_cpc..." is often a bit too intense...

t-pi:
...need one addition for get_status()

multigeiger/display.cpp
```diff
@@ -28,25 +28,28 @@ void display_start_screen(void) {
   pu8x8->clear();
   if (isLoraBoard) {
     // Display is only 4 lines by 8 characters; lines counting from 2 to 5
-    pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);        // u8x8 can handle only 8x8, smaller font has no advantage
+    pu8x8->setFont(u8x8_font_victoriamedium8_r);
     for (int i = 2; i < 6; i++) {
       pu8x8->drawString(0, i, "        ");  // clear all 4 lines
     }
```
ThomasWaldmann, line 34: as you have a lora board, can you check whether these 3 lines are really needed?

see line 28, that should already clear the display.

multigeiger/display.cpp
```diff
@@ -28,25 +28,28 @@ void display_start_screen(void) {
...
+    pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
+    pu8x8->drawString(0, 2, "Multi-");
+    pu8x8->drawString(0, 3, " Geiger");
+    pu8x8->setFont(u8x8_font_victoriamedium8_r);
     snprintf(line, 9, "%s", VERSION_STR);  // 8 chars + \0 termination
-    pu8x8->drawString(0, 5, line);
+    pu8x8->drawString(0, 4, line);
   } else {
     pu8x8->setFont(u8x8_font_amstrad_cpc_extended_f);
-    pu8x8->drawString(0, 0, "Geiger-Counter");
```
ThomasWaldmann, line 45: I'ld prefer to just change this to `Multi-Geiger` or `MultiGeiger` and keep the underlining.

In any case, not `Multi-Geiger-Counter`.

multigeiger/display.cpp
```diff
@@ -91,20 +94,27 @@ static const char *status_chars[STATUS_MAX] = {
...
+      status[index] = value;
+      displayStatus();
+    }
+  } else
     log(ERROR, "invalid parameters: set_status(%d, %d)", index, value);
 }
 
+int get_status(int index) {
+  return status[index];
+}
```
ThomasWaldmann, line 116: do you use that already?

multigeiger/display.cpp
ThomasWaldmann: that is what #208 was about, iirc.

ThomasWaldmann: and if they actually **are** needed, we have a clearLine (or so) function that should be called for this.

rexfue:
New font looks good.

Some suggestions:
- big display:
  - center Multi-Geiger and keep underlining (as TW suggested)
  - replace Info:boehri.de with ecocurious.de  (@WwwBoehriDe: what do you think?)
- small display:
  - write 'Geiger' (second line)  either left or right justified 
  - replace the right dash at the version number with a 'd' for 'develop'

ThomasWaldmann:
It points to a visually appealing website, but there is not much information there (and also, the link to github is broken, misses the "2").

For boehri.de, it is kind of the opposite. :-)

multigeiger/display.cpp
t-pi: ok, will revert

multigeiger/display.cpp
t-pi: here, no. But in ble PR. As it is quite generic, I put it here together with the other display related commits.

t-pi:
Maybe @WwwBoehriDe could make a landing page, e.g. boehri.de/mg, with links to github, ecocurious and multigeiger.citysensor.de? Neither on boehri.de nor on ecocurious (currently) there is a quick pointer to here...

t-pi:
@rexfue, the version number comes from VERSION_STR, so I prefer not to tamper with it...

multigeiger/display.cpp
t-pi: No, works fine without the extra lines. Removed them.

t-pi:
Hmm, I'm getting this error upon initialization on both boards:
```
...
Loaded config 'send2Madavi'= '1'
E (728) MCPWM: /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/driver/mcpwm.c:190 (mcpwm_set_duty_type):MCPWM DUTY TYPE ERROR
E (740) MCPWM: /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/driver/mcpwm.c:190 (mcpwm_set_duty_type):MCPWM DUTY TYPE ERROR
```
Did you notice?

ThomasWaldmann:
@t-pi yes, seeing these 2 also. but i didn't see any erroneous calls / parameters in the source.

multigeiger/display.cpp
```diff
@@ -64,18 +61,18 @@ void setup_display(bool loraHardware) {
   display_start_screen();
 }
 
-void clearDisplayLine(int line) {
+void clear_displayline(int line) {
```
ThomasWaldmann, line 64: scope creep again.

a better way to deal with such stuff is to resist scope creep in a PR and just file a ticket for a style refactoring.
and then do that in a separate PR solving it globally.

t-pi:
quick question, would mg.citysensor.de as link be doable? But then further links from the map to github / ecocurious would be helpful

ThomasWaldmann:
Guess we just keep the link as is for now. If we want to change the URL, we first need a new pretty and short url, some reasonable content there and *then* we'll change the MG firmware.

ThomasWaldmann:
@t-pi do you want to do a `git rebase -i master` and collapse some of the changes together (but maybe keep the style refactor separate?)?

From the github UI, i could only squash *all* together.

multigeiger/display.cpp
t-pi: Guilty as charged. I just wanted to change the local functions in display.cpp but got indeed carried away. 
Sorry, will avoid this next time!

t-pi:
Okay, did a git rebase -i master.
Hope that's okay like this?

ThomasWaldmann:
That does not look like a rebase. Did you push (usually you need push -f)?

t-pi:
Re-tried with -f, but did not change the outcome. Trying to re-rebase started to get confusing (lot's of conflicts that pointed to mixing of commits), so I preferred to abort... 🙄

ThomasWaldmann:
I can try to help with the rebase, just do not change the PR until I am finished.

t-pi:
Ok, thanks! Tell me if what I can do to help!

ThomasWaldmann:
Working on the rebase -i now...

ThomasWaldmann:
somehow the changes were duplicate in this PR:
- once in original form + astyle fixup commits
- another incarnation with same content but astyle fixups squashed into the preceeding commits

so somehow the rebase went very wrong. usually it is just `git rebase -i master` from the feature branch you're working on (and a current master branch).

to fix the duplication, I just dropped the original form of the changes and kept the squashed form.

then i edited the commit comment and removed the astyle part (which is not interesting after collapsing).

have a look, but i guess it's now in a good state (besides that stylistic/cosmetic changes were done together with functional changes, which is not easy to dissect any more, so just don't do that in future).

t-pi:
Ok for me. Thanks!
What I did was a git rebase -i master, and then I marked all for squash except the first commit and the one with the snake_case... Probably that got it all mixed up, sorry! Will avoid commit creep in the future indeed...

PR #318, opened 2020-05-21
Remove version numbers / dates from filenames

### LINKED ISSUE #204: remove version numbers and dates from filenames
this is one of the changes done by #156.

Makes sense, we should never have dates or version numbers in filenames as this breaks git history for that file. We don't need dates / version numbers in the filenames as we use git for history / versioning / revision control.

Inside the file, a date or version number can be mentioned (just should not be forgotten when updating it).

ThomasWaldmann:
fixes #204.

i did not change file contents, just the file names.

i commited them in natural order (earliest version first, current version last).

in future, it would be good to commit corresponding versions of files at once and optionally use `git tag` to tag them with some version number, like `hw1.4` when referring to that hw version.

and of course, do not add dates or version numbers (again) to filenames, use git for versioning.


rexfue:
Checked board and schematic files -> OK

PR #321, opened 2020-05-21
updated de translation

ThomasWaldmann:


ThomasWaldmann:
needed to test readthedocs updating / translating.

PR #326, opened 2020-05-22
added comment about CONFIG_VERSION

rexfue:


multigeiger/webconf.cpp
```diff
@@ -63,7 +63,11 @@ void format_bool(bool *value, char *text) {
   strcpy(text, *value ? "1" : "0");
 }
 
-#define CONFIG_VERSION "013"  // for IoTWebConfig
+// This only needs to be changed if the layout of the configuration is changed. 
+// Appending new variables does not require a new version number here.
+// If this value is changed, ALL configuration variables must be re-entered, 
+// including the WiFi credentials.
+#define CONFIG_VERSION "013"
```
ThomasWaldmann, line 70: Sounds good. I guess astyle is unhappy about some trailing blanks?

PR #333, opened 2020-05-22
do async NTP/clock setup, fixes #316

### LINKED ISSUE #316: no wifi -> ntp setup blocks. no geiger counting.
I just did an update from 1.12 to 1.15-dev on a workshop device that has not been set up, and it was stuck in the welcome screen. That is not a great user experience, especially at the workshops, where there might be no Wifi...

I guess this comes from the new NTP functionality:
```
  // a bug in arduino-esp32 1.0.4 crashes the esp32 if the wifi is not
  // configured yet and one tries to configure for NTP:
  while (iotWebConf.getState() <= IOTWEBCONF_STATE_NOT_CONFIGURED) {
    iotWebConf.delay(200);
  }
  setup_clock(0);  // 0 == do NTP!
```
Could we somehow async the NTP setup?

ThomasWaldmann:
works for me.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -71,19 +71,30 @@ void setup() {
   setup_speaker(playSound, ledTick && switches.led_on, speakerTick && switches.speaker_on);
   setup_transmission(VERSION_STR, ssid, isLoraBoard);
   setup_ble(ssid, sendToBle && switches.ble_on);
+  setup_log_data(SERIAL_DEBUG);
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Well, i can't set that bool there as it'll take a while after setup until it timesyncs.

But IIRC (after implementing that all), I found that the TLS stuff on ESP32 might not check for expired certs anyway.

So guess we can add a bool IF we run into a problem.

PR #344, opened 2020-05-23
docs: document dip switches and buttons

### LINKED ISSUE #128: Readme: DIP-Schalter
Beschreibung des optionale DIP-Schalters. Bestückung und  Funktion

ThomasWaldmann:


docs/source/usage.rst
```diff
@@ -75,6 +75,17 @@ Positions:
...
+ESP32 buttons
+#############
+
+The MultiGeiger ESP32 microcontroller board has 2 buttons:
+
+- RST: reset (restarts device)
+- PRG: program (press and hold PRG, press RST momentarily, upload firmware)
+
+Often, using PRG is not needed for flashing firmware, because it "just works" when using the right tools.
+
```
t-pi, line 87: + recommendation / link to OTA FW update via WiFi?

docs/source/usage.rst
```diff
@@ -75,6 +75,17 @@ Positions:
...
+
 DIP Switches
 ############
```
t-pi, line 91: Is it possible to generate a table of content? With OLED, BLE, buttons and switches it gets longer, and a toc at the top would definitively help.

docs/source/usage.rst
ThomasWaldmann: there should be a chapter about updating the device.

but we do not have that yet, thus we can't link to it.

docs/source/usage.rst
ThomasWaldmann: There is a TOC at the left side.

ThomasWaldmann:
General question: are the usual dip switches (electronic parts) really labelled 0..3? Or rather 1..4?

The source code referred to SW0 .. SW3 so I just took that, wondering...

rexfue:
the labels on the hardware switches are 1..4

ThomasWaldmann:
@rexfue thanks for the info, I created #349 for this. not in scope of this PR.

PR #347, opened 2020-05-23
docs: add deployment guide (en), rst format

ThomasWaldmann:
looks like this when the rendered html file is directly viewed in the browser:

[deployment.pdf](https://github.com/ecocurious2/MultiGeiger/files/4672567/deployment.pdf)

on readthedocs, there would be also some navigation at the left side (and maybe other theme additions).

Issues / TODO:

multigeiger image needs to get A B C labels

garden plan image should also have A B C ... labels rather than german text inside the image (we can translate in rst text part, but we can not translate inside an image).

proof reading by native speaker


ThomasWaldmann:
Also, the table markup is a bit ugly on github diff display.

I used it to place the text to the left of the multigeiger picture.

ThomasWaldmann:
@WwwBoehriDe ^^^

WwwBoehriDe:
Cool to have the english document.

ThomasWaldmann:
@WwwBoehriDe after the german strings are entered on transifex, it will generate the german document from same source.

ThomasWaldmann:
Considering that Marianne works on these docs, too, I merged this now "as is" (content-wise) so there won't be blocking or double work.

If there is any content that needs changing here, it can be done later.

PR #374, opened 2021-02-18
manual translated

Baitcookie:


ThomasWaldmann:
Considering that Marianne works on these docs, too, I'll clean this up a bit (resolve conflict, squash commits) and merge this now "as is" (content-wise) so there won't be conflicts or blocking.

If there is any content that needs changing here, it can be done later.

ThomasWaldmann:
@Baitcookie thanks for helping!

PR #412, opened 2021-04-02
Changed switch numbering,  fixes #349

### LINKED ISSUE #349: inconsistent switch numbering
the dip switch contacts are numbered 1 2 3 4 on the hardware part (this is what the end user sees).

on the schematic and in the software, they are numbered SW0 SW1 SW2 SW3.

the log output also refers to SW0 SW1 SW2 SW3.

to get the numbers consistent, the schematic has to be changed and afterwards the software and the docs.

please also check whether SW0 == 1, SW1 == 2, SW2 == 3, SW3 == 4, so that we get this really right.

rexfue:


ThomasWaldmann:
Notable: 4 3 2 1 switch numbering and "off" at the bottom

Isn't that usually 1 2 3 4 and "off" at the top?

Is the symbol for the switch part rotated on the schematic and should it be turned 180 degrees?

rexfue:
There is no "Off" on the schematic, it is "ON".  I think, up and down, top or bottom is irrelevant for the schematic, important is the numbering on the switch itself.

ThomasWaldmann:
I think it would help if the schematic part looks correct (like when looking on the physical part), esp. if it matters which switch you want to switch (whether leftmost is 1 or is 4).

PR #413, opened 2021-04-24
adapted to iotwebconf 3.0

rexfue:


ThomasWaldmann:
if the user did not enter values yet (like after flashing the new firmware), does it default to the settings in `userdefines.h` and initialize the strings using these values?

ThomasWaldmann:
did you try OTA update already?

rexfue:
not yet. will check.

rexfue:
OK, tested both - they work.

ThomasWaldmann:
Looks like there is a conflict now. Intend to merge this **after** 1.15.1.

Did you test with iotwebconf 3.0.2?

rexfue:
I don't understand, why there is still the conflict. I pulled the master, rebased the branch to master and pushed it.
Btw. I have tested with IotWebConf 3.0.2 -> everything works as expected on both hardwares , also OTA works.

rexfue:
Made an new PR to avoid merge conflict

PR #414, opened 2021-05-02
Fix Adafruit BME680 lib V2.0 #macro name change

t-pi:
BME680 I2C address defines were changed in 'bme68x_defs.h' from Adafruit BME680 lib V2.0.0 to:
BME68X_I2C_ADDR_LOW / _HIGH
instead of _PRIMARY / _SECONDARY

Fixed here.

### DIFF
--- a/multigeiger/thp_sensor.cpp
+++ b/multigeiger/thp_sensor.cpp
@@ -23,9 +23,9 @@ bool setup_thp_sensor(void) {
 
   // BME680
   if (type_thp == 0) {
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
https://github.com/ecocurious2/MultiGeiger/blob/master/platformio-example.ini#L28

should that line be updated to >=2.0 or so, because this change breaks all older versions?

t-pi:
Good point, added @^2.0.0 to the Adafruit BME680 lib_dep

PR #415, opened 2021-05-03
BLE via NimBLE-Arduino lib

t-pi:
Swapped BLE management from Arduino to [NimBLE](https://github.com/h2zero/NimBLE-Arduino) to save RAM and flash.
Changes running locally for a while already under 1.15-dev. Now ported to 1.16-dev.
Includes changes required for Adafruit BME680 V2.0.0.

Before:
RAM:   [==        ]  20.4% (used 66744 bytes from 327680 bytes) 
Flash: [=====     ]  52.4% (used 1752965 bytes from 3342336 bytes)

After:
RAM:   [==        ]  18.9% (used 62032 bytes from 327680 bytes) 
Flash: [====      ]  40.1% (used 1340817 bytes from 3342336 bytes)

### DIFF
--- a/multigeiger/ble.cpp
+++ b/multigeiger/ble.cpp
@@ -14,23 +14,15 @@
 #include "ble.h"
 #include "display.h"
 
-#include <BLEDevice.h>
-#include <BLEServer.h>
-#include <BLEUtils.h>
-#include <BLE2902.h>
+#include <NimBLEDevice.h>
+
+static NimBLEServer *bleServer;
 
 #define BLE_SERVICE_HEART_RATE    BLEUUID((uint16_t)0x180D)  // 16 bit UUID of Heart Rate Service
 #define BLE_CHAR_HR_MEASUREMENT   BLEUUID((uint16_t)0x2A37)  // 16 bit UUID of Heart Rate Measurement Characteristic
 #define BLE_CHAR_HR_POSITION      BLEUUID((uint16_t)0x2A38)  // 16 bit UUID of Heart Rate Sensor Position Characteristic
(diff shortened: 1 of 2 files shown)


t-pi:
@ThomasWaldmann, it's been a while, where do I find the astyle detail issues? The "Details" above only show an empty "Error:" line.
Thanks!

t-pi:
Ah, @TW, sorry, never mind - NoScript issue with github.useractions.whateversubdomain... 🙄

t-pi:
Why can't astyle run through the whole file at once 😑

multigeiger/thp_sensor.cpp
```diff
@@ -23,9 +23,9 @@ bool setup_thp_sensor(void) {
 
   // BME680
   if (type_thp == 0) {
-    if (bme680.begin(BME680_I2C_ADDR_PRIMARY))
+    if (bme680.begin(BME68X_I2C_ADDR_LOW))
       type_thp = 680;
-    else if (bme680.begin(BME680_I2C_ADDR_SECONDARY))
+    else if (bme680.begin(BME68X_I2C_ADDR_HIGH))
       type_thp = 680;
   }
```
ThomasWaldmann, line 30: this does not belong into this PR - can you rebase it onto current master?

t-pi:
Dang, that doesn't look right... yet!

t-pi:
Tested and runs stable now with both MultiGeigers here: 1x Wifi Kit, 1x Lora Kit (only Wifi used, though)

rexfue:
Testet also both Multigeiger: works!

PR #417, opened 2021-05-06
Fix sensor.community domain

MarcelWaldvogel:
The config page and the docs referred to `sensors.community`, which is a different service.


ThomasWaldmann:
Good catch, thanks for the PR!

Issue #416, opened 2021-05-04
Add simple buzzer alarm based on historic and current values


t-pi:
We've discussed this during the meeting already.
One option is an email via a script checking the online data - not fully in focus here.
Another option is a buzzer alarm, if some thresholds are reached, e.g.

1. long term averaged value > 500nSv/h?
2. (short term values over e.g. 1 min) > (long term values * 3)

With the interrupt based approach from a recent version,  @ThomasWaldmann, could we play an alarm sound while keeping the system running? In that case options (1) and (2) could be easily integrated into the firmware, e.g. as #define options.

Thomas, @rxf, @WwwBoehriDe, what do you think about the thresholds?

ThomasWaldmann:
The code has a little sequencer-like tone playing system, so technically i don't see a problem (you just won't hear ticks and tones at the same time).

But **if** a warning functionality is wanted, something server based / based on the online data would be more useful. I don't think we want to send email or other notifications directly from the multigeiger.

Also, a buzzer alarm is only useful if you are close to the device, which is usually not the case for the current outside deployment method.

BTW, there has already been some discussion somewhere whether one wants to send out "alarm notifications", especially if that functionality is rather public. That could be useful but it could also cause a panic just because of some malfunction.

t-pi:
Indeed, my local code works with the sequencer, after a few small modifications. Did add this on top of my NimBLE PR, so will push this after merge.

Regarding the local alarm being of no use, I disagree. 

We don't see radiation, and after a few days I rarely look up the values online. Of course a mail alarm will help, and if IFTTT works as discussed last Thursday, this can also be implemented quickly without the need to own a server to have the mail script running. However my MultiGeiger sits in my flat, so an alarm would be much more immediate. The one at my parents' garden would also be heard while staying (or passing) outside the house - and probably all Geigers are sitting in a few meters distance from a wall plug, so they should be heard, when in proximity.

So I implemented it as a firmware compilation option as of now. Whoever wants this, must thus activate the compile option in ```userdefines.h``` and can alter the value (0.5µSv/h accumulated or 3x accumulated). The alarm sequence will play every 10s, when ```publish``` is called. You can then try it out and listen for yourself if this could work - after merging NimBLE 😄

ThomasWaldmann:
OK, it it is useful to you, go on. I just imagined my usecase (when deployed to garden, i usually won't hear it - maybe with a delay of days which would likely mean that i get something from the internet news first).

[...]

PR #427, opened 2021-05-15
Local alarm compile

t-pi:
- Some optimizations to the sound sequencer
- Added alarm sequence & function
- Added userdefines.h parameters for enable and threshold
- Added IotWebConf parameters for enable and threshold
- Alarm will sound once every 10 s (every ```publish```) if 
 - accumulated dose rate > set threshold (default 0.5 µSv/h)
 - current dose rate > 3x accumulated dose rate

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -148,6 +148,17 @@ void publish(unsigned long current_ms, unsigned long current_counts, unsigned lo
     display_GMC(((int)accumulated_time / 1000), (int)(accumulated_Dose_Rate * 1000), (int)(Count_Rate * 60),
                 (showDisplay && switches.display_on));
 
+    // Sound local alarm?
+    if (soundLocalAlarm && GMC_factor_uSvph > 0) {
+      if (accumulated_Dose_Rate > localAlarmThreshold) {
+        log(INFO, "Local alarm: Accumulated dose of %.3f µSv/h above threshold at %.3f µSv/h", accumulated_Dose_Rate, localAlarmThreshold);
+        alarm();
+      } else if (Dose_Rate > (accumulated_Dose_Rate * localAlarmFactor)) {
+        log(INFO, "Local alarm: Current dose of %.3f > %d x accumulated dose of %.3f µSv/h", Dose_Rate, localAlarmFactor, accumulated_Dose_Rate);
+        alarm();
+      }
+    }
+
     if (Serial_Print_Mode == Serial_Logging) {
       log_data(counts, dt, Count_Rate, Dose_Rate, hv_pulses,
                accumulated_GMC_counts, accumulated_time, accumulated_Count_Rate, accumulated_Dose_Rate,
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -151,10 +151,10 @@ void publish(unsigned long current_ms, unsigned long current_counts, unsigned lo
     // Sound local alarm?
     if (soundLocalAlarm && GMC_factor_uSvph > 0) {
       if (accumulated_Dose_Rate > localAlarmThreshold) {
-        log(INFO, "Local alarm: Accumulated dose of %.3f µSv/h above threshold at %.3f µSv/h", accumulated_Dose_Rate, localAlarmThreshold);
+        log(WARNING, "Local alarm: Accumulated dose of %.3f µSv/h above threshold at %.3f µSv/h", accumulated_Dose_Rate, localAlarmThreshold);
         alarm();
       } else if (Dose_Rate > (accumulated_Dose_Rate * localAlarmFactor)) {
-        log(INFO, "Local alarm: Current dose of %.3f > %d x accumulated dose of %.3f µSv/h", Dose_Rate, localAlarmFactor, accumulated_Dose_Rate);
+        log(WARNING, "Local alarm: Current dose of %.3f > %d x accumulated dose of %.3f µSv/h", Dose_Rate, localAlarmFactor, accumulated_Dose_Rate);
         alarm();
       }
     }
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -22,6 +22,7 @@ volatile int *isr_tick_sequence = NULL;
 volatile int *isr_sequence = NULL;  // currently played sequence
 
 static int tick_sequence[8];
+static int alarm_sequence[12];
 
 // hw timer period and microseconds -> periods conversion
 #define PERIOD_DURATION_US 1000
@@ -70,7 +71,6 @@ void IRAM_ATTR isr_audio() {
 
   if (!p)
     return;  // nothing to do
-  // do not access *p below here, p might point to uninitialized memory after the sequence array!
 
   // note: by all means, **AVOID** mcpwm_set_duty() in ISR, causes floating point coprocessor troubles!
   //       when just calling mcpwm_set_duty_**type**(), it will reuse a previously set duty cycle.
@@ -123,7 +123,8 @@ void IRAM_ATTR tick(bool high) {
   // high true: "tick" -> high frequency tick and LED blink
   // high false: "tock" -> lower frequency tock, no LED
   // called from ISR!
-  portENTER_CRITICAL_ISR(&mux_audio);
+  if (isr_sequence)
+    return;
 
   int *sequence;
 
@@ -142,8 +143,9 @@ void IRAM_ATTR tick(bool high) {
     sequence[6] = led_tick ? (high ? 0 : -1) : -1;
     sequence[7] = 0;  // END
   } else
-    sequence = NULL;
+    return;
 
+  portENTER_CRITICAL_ISR(&mux_audio);
   isr_tick_sequence = sequence;
   portEXIT_CRITICAL_ISR(&mux_audio);
 }
@@ -158,6 +160,33 @@ void tick_enable(bool enable) {
     led_tick = false;
     speaker_tick = false;
   }
+}
+
+void alarm() {
+
+  int *sequence;
+
+  sequence = alarm_sequence;
+  // "high_pitch"
+  sequence[0] = 3000000;  // frequency_mHz
+  sequence[1] = 1;  // volume
+  sequence[2] = -1;  // LED, -1 = don't touch
+  sequence[3] = 400;  // duration_ms
+  // "low_Pitch"
+  sequence[4] = 1000000;
+  sequence[5] = 1;
+  sequence[6] = -1;
+  sequence[7] = 400;
+  // "off"
+  sequence[8] = 0;
+  sequence[9] = 0;
+  sequence[10] = -1;
+  sequence[11] = 0;  // END
+
+  // called from ISR!
+  portENTER_CRITICAL_ISR(&mux_audio);
+  isr_audio_sequence = sequence;
+  portEXIT_CRITICAL_ISR(&mux_audio);
 }
 
 void play(int *sequence) {
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -24,11 +24,11 @@ volatile int *isr_sequence = NULL;  // currently played sequence
 static int tick_sequence[8];
 static int alarm_sequence[12] = {
   // "high_Pitch"
-  3000000, 1, -1, 400,  // frequency_mHz, volume, LED, -1 = don't touch, duration_ms
+  3000000, 1, -1, 400,  // frequency_mHz, volume, LED (-1 = don't touch), duration_ms
   // "low_Pitch"
   1000000, 1, -1, 400,
   // "off"
-  0, 0, -1, 0 // duration = 0 --> END
+  0, 0, -1, 0 // duration_ms = 0 --> END
 };
 
 
@@ -170,9 +170,10 @@ void tick_enable(bool enable) {
 }
 
 void alarm() {
-  portENTER_CRITICAL_ISR(&mux_audio);
+  // play alarm sound, called from normal code (not ISR)
+  portENTER_CRITICAL(&mux_audio);
   isr_audio_sequence = alarm_sequence;
-  portEXIT_CRITICAL_ISR(&mux_audio);
+  portEXIT_CRITICAL(&mux_audio);
 }
 
 void play(int *sequence) {
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -183,7 +183,6 @@ void alarm() {
   sequence[10] = -1;
   sequence[11] = 0;  // END
 
-  // called from ISR!
   portENTER_CRITICAL_ISR(&mux_audio);
   isr_audio_sequence = sequence;
   portEXIT_CRITICAL_ISR(&mux_audio);
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -79,6 +79,7 @@ void IRAM_ATTR isr_audio() {
 
   if (!p)
     return;  // nothing to do
+  // do not access *p below here, p might point to uninitialized memory after the sequence array!
 
   // note: by all means, **AVOID** mcpwm_set_duty() in ISR, causes floating point coprocessor troubles!
   //       when just calling mcpwm_set_duty_**type**(), it will reuse a previously set duty cycle.
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -22,7 +22,15 @@ volatile int *isr_tick_sequence = NULL;
 volatile int *isr_sequence = NULL;  // currently played sequence
 
 static int tick_sequence[8];
-static int alarm_sequence[12];
+static int alarm_sequence[12] = {
+  // "high_Pitch"
+  3000000, 1, -1, 400,  // frequency_mHz, volume, LED, -1 = don't touch, duration_ms
+  // "low_Pitch"
+  1000000, 1, -1, 400,
+  // "off"
+  0, 0, -1, 0 // duration = 0 --> END
+};
+
 
 // hw timer period and microseconds -> periods conversion
 #define PERIOD_DURATION_US 1000
@@ -123,8 +131,7 @@ void IRAM_ATTR tick(bool high) {
   // high true: "tick" -> high frequency tick and LED blink
   // high false: "tock" -> lower frequency tock, no LED
   // called from ISR!
-  if (isr_sequence)
-    return;
+  portENTER_CRITICAL_ISR(&mux_audio);
 
   int *sequence;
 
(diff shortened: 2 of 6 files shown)


t-pi:
Adresses #416

multigeiger/speaker.cpp
```diff
@@ -58,17 +59,18 @@ void IRAM_ATTR isr_audio() {
       playing_tick = true;
     }
   }
+  portEXIT_CRITICAL_ISR(&mux_audio);
+
```
ThomasWaldmann, line 63: iirc, isr_{sequence,tick_sequence,audio_sequence} are the global isr variables protected by mux_audio.

the code below is accessing isr_sequence, so it should be in the protected block.

multigeiger/userdefines-example.h
```diff
@@ -41,6 +41,12 @@
 // Play a start sound at boot/reboot time?
 #define PLAY_SOUND true
 
+// Play an alarm sound when radiation level too high?
```
ThomasWaldmann, line 44: ... is too high?

multigeiger/multigeiger.ino
```diff
@@ -147,6 +147,14 @@ void publish(unsigned long current_ms, unsigned long current_counts, unsigned lo
     display_GMC(((int)accumulated_time / 1000), (int)(accumulated_Dose_Rate * 1000), (int)(Count_Rate * 60),
                 (showDisplay && switches.display_on));
 
+    // Sound local alarm?
+    if (soundLocalAlarm && GMC_factor_uSvph > 0) {
+      if ((accumulated_Dose_Rate > localAlarmThreshold) || (Dose_Rate > accumulated_Dose_Rate * 3)) {
+        log(INFO, "Local alarm: current dose %.3f vs. accumulated dose %.3f µSv/h", Dose_Rate, accumulated_Dose_Rate);
+        alarm();
+      }
```
ThomasWaldmann, line 155: guess it would be good to have 2 IFs, one for each term and also to log a specific message for each condition, like:

1. "Local alarm: accumulated dose %.3f µSv/h > alarm threshold"
2. "Local alarm: current dose %.3f > 3 * accumulated dose %.3f µSv/h"

ThomasWaldmann: and guess the 3 also needs to be configurable?

multigeiger/speaker.cpp
```diff
@@ -156,6 +163,33 @@ void tick_enable(bool enable) {
   }
 }
 
+void IRAM_ATTR alarm() {
```
ThomasWaldmann, line 166: why is this IRAM_ATTR?

multigeiger/speaker.cpp
```diff
@@ -156,6 +163,33 @@ void tick_enable(bool enable) {
...
+  // "off"
+  sequence[8] = 0;
+  sequence[9] = 0;
+  sequence[10] = -1;
+  sequence[11] = 0;  // END
+
+  // called from ISR!
+  portENTER_CRITICAL_ISR(&mux_audio);
+  isr_audio_sequence = sequence;
+  portEXIT_CRITICAL_ISR(&mux_audio);
```
ThomasWaldmann, line 190: you are calling this from main loop, not from ISR.

ThomasWaldmann (commented):
this is mixing a fix/optimization(?) to existing hard-to-get-right code (the sequencer isr / mutex) with a new feature in the same commit. please avoid.

multigeiger/speaker.cpp
t-pi: My understanding was that only writing to the critical variables should be protected. Must also the reading be protected?

multigeiger/speaker.cpp
t-pi: Ah, thought that anything that plays with the &mux_audio must be quickly accessible from IRAM. But e.g. `play` is not, so probably not necessary. Removed in next commit.

multigeiger/speaker.cpp
ThomasWaldmann: a mutex means that the protected blocks are mutually exclusive.

if writing is an issue and needs a mutex (likely because it is not atomic somehow), i guess one also must not read while a write is in progress. and if the read should be mutually excluded, it also needs to be in a mutex block.

t-pi: Okay, will move the mutex back

multigeiger/userdefines-example.h
t-pi: Added some further text and moved the alarm block to the end, as in IotWebConf

multigeiger/multigeiger.ino
t-pi: Done

multigeiger/speaker.cpp
ThomasWaldmann: also, I guess the mutex related calls also have quite a bit of overhead, so just keeping the mutex for some simple C operations might be more efficient than releasing and re-acquiring it more frequently.

multigeiger/speaker.cpp
t-pi: Comment removed

ThomasWaldmann:
can you please update your master branch and rebase this onto the current master?

multigeiger/speaker.cpp
t-pi: Right, makes sense. Kept the mutex active for the `!isr_sequence` poll though.

t-pi:
> 
> this is mixing a fix/optimization(?) to existing hard-to-get-right code (the sequencer isr / mutex) with a new feature in the same commit. please avoid.

Question: How can I work on basically two branches at the same time? E.g. for the alarm I needed to work on speaker, that's why this got mixed up. Is there a preferred way to work on "lower level" commits in one branch and "higher level" commits in a second one?

t-pi:
Resolves #416

ThomasWaldmann:
You could do 2 branches and (re)base the feature branch onto the bugfix branch now and then.

master -> bugfix -> feature

Of course, there could be conflicts now and then depending on how changes overlap.

Or just do the same within 1 branch, but keep fixing and adding stuff in separate commits always.
If you need multiple commits for each, you can later use `git rebase -i` and s(quash) or f(ixup) to merge some commits again,
e.g squash fix1 fix2 fix3 commits, squash feature1 feature2 feature2 commits, so you'll end up with 1 fix commit and 1 feature commit.

ThomasWaldmann:
I am still having troubles to review this / to see why the ISR code got changes / whether they improve the code.

So, may I suggest:

- don't tell "some optimizations", but specifically which / why
- don't tell "adjusted mux_audio back", but specifically tell what you changed and why you changed it.
- differentiate between bug fixes (if you found a bug, tell why it is a bug) and other changes (also tell why)

multigeiger/speaker.cpp
```diff
@@ -58,19 +59,20 @@ void IRAM_ATTR isr_audio() {
...
     duration_ms = *p++;
-    isr_sequence = p;
+    isr_sequence = duration_ms == 0 ? NULL : p;
+  } else {
+    portEXIT_CRITICAL_ISR(&mux_audio);
+    return;  // nothing to do
   }
   portEXIT_CRITICAL_ISR(&mux_audio);
 
-  if (!isr_sequence)
```
t-pi, line 71: @ThomasWaldmann, here we accessed isr_sequence after exiting the &mux_audio. In the PR this is prevented. Will create a separate PR once the code compiles with IotWebConf 3.1.x.

multigeiger/speaker.cpp
```diff
@@ -106,10 +108,13 @@ void IRAM_ATTR isr_audio() {
     // duration == 0 marks the end of the sequence to play
     portENTER_CRITICAL_ISR(&mux_audio);
     isr_sequence = NULL;
-    if (playing_tick)
+    if (playing_tick) {
       isr_tick_sequence = NULL;
-    else if (playing_audio)
+      playing_tick = false;
```
t-pi, line 113: ... and the second bugfix is that `playing_tick `and `playing_audio` were never reset to false, being declared as static. Thus the reset here. Will also be part of the new bugfix-PR, after IotWebConf-Fix, @ThomasWaldmann.

multigeiger/speaker.cpp
ThomasWaldmann: one could avoid having the portEXIT_CRITICAL_ISR(&mux_audio) duplicated to the else block by setting p = NULL there and afterwards just checking the value of p (local, instead of accessing global volatile isr_sequence).

smaller diff and easier mux ops.

ThomasWaldmann: or rather:
```
  volatile int *p = isr_sequence;
  if (p) {
        ...
        isr_sequence = duration_ms == 0 ? NULL : p;  
  }
  portEXIT_CRITICAL_ISR(&mux_audio);

  if (!p)
    return;  // nothing to do
```

multigeiger/speaker.cpp
```diff
@@ -119,7 +124,8 @@ void IRAM_ATTR tick(bool high) {
   // high true: "tick" -> high frequency tick and LED blink
   // high false: "tock" -> lower frequency tock, no LED
   // called from ISR!
-  portENTER_CRITICAL_ISR(&mux_audio);
+  if (isr_sequence)
+    return;
```
ThomasWaldmann, line 128: 2 issues here:
- why do you check this?
- when accessing this, you need to have the mutex which you just removed from above.

multigeiger/speaker.cpp
```diff
@@ -138,8 +144,9 @@ void IRAM_ATTR tick(bool high) {
     sequence[6] = led_tick ? (high ? 0 : -1) : -1;
     sequence[7] = 0;  // END
   } else
-    sequence = NULL;
+    return;
```
ThomasWaldmann, line 147: yeah, resetting isr_tick_sequence to NULL is done already at the end of playing the tick sound, so this is not required here.

i somewhat dislike returning from the else-block though.
if this change is kept, maybe rather move the block below into the if-block?

multigeiger/speaker.cpp
ThomasWaldmann: oh, oops, yeah - confirmed. X-)

multigeiger/speaker.cpp
t-pi: Ok, adapted your proposition. Separate PR in preparation.

multigeiger/speaker.cpp
t-pi: I changed the static variables to normal disposable ones. That feels more straightforward.

multigeiger/speaker.cpp
t-pi: This is in a different function I didn't touch. I initally also fell for the hidden lines in Github's diff view  :)

multigeiger/speaker.cpp
t-pi: Alright, moved the assignment block into the if block

multigeiger/speaker.cpp
ThomasWaldmann: thanks!

multigeiger/speaker.cpp
ThomasWaldmann: \o/

multigeiger/speaker.cpp
ThomasWaldmann: see the other PR.

multigeiger/speaker.cpp
ThomasWaldmann: i checked master branch: the tick function does not have that "if" in master branch.

your branch has it, so not sure what you mean.

t-pi: obsolete

multigeiger/multigeiger.ino
```diff
@@ -148,6 +148,17 @@ void publish(unsigned long current_ms, unsigned long current_counts, unsigned lo
...
+    // Sound local alarm?
+    if (soundLocalAlarm && GMC_factor_uSvph > 0) {
+      if (accumulated_Dose_Rate > localAlarmThreshold) {
+        log(INFO, "Local alarm: Accumulated dose of %.3f µSv/h above threshold at %.3f µSv/h", accumulated_Dose_Rate, localAlarmThreshold);
+        alarm();
+      } else if (Dose_Rate > (accumulated_Dose_Rate * localAlarmFactor)) {
+        log(INFO, "Local alarm: Current dose of %.3f > %d x accumulated dose of %.3f µSv/h", Dose_Rate, localAlarmFactor, accumulated_Dose_Rate);
+        alarm();
+      }
+    }
```
ThomasWaldmann, line 160: just thought whether this should use log at WARNING or even CRITICAL level rather.

multigeiger/speaker.cpp
```diff
@@ -22,6 +22,15 @@ volatile int *isr_tick_sequence = NULL;
 volatile int *isr_sequence = NULL;  // currently played sequence
 
 static int tick_sequence[8];
+static int alarm_sequence[12] = {
+  // "high_Pitch"
+  3000000, 1, -1, 400,  // frequency_mHz, volume, LED, -1 = don't touch, duration_ms
```
ThomasWaldmann, line 27: maybe rather:

// frequency_mHz, volume, LED (-1 = don't touch), duration_ms

if this is badly phrased at other places also, this could be a separate cleanup PR.

multigeiger/userdefines-example.h
```diff
@@ -63,3 +63,20 @@
 // 0x2A38: Heart Rate Sensor Position --> sends TUBE_TYPE
 // 0x2A39: Heart Rate Control Point --> allows to reset "energy expenditure", as required by service definition
 #define SEND2BLE false
+
+// Play an alarm sound when radiation level is too high?
+// Activates when either accumulated dose rate reaches the set threshold (s. below)
+// or when the current dose rate is higher than the accumulated dose rate by the set factor (s. below)
```
ThomasWaldmann, line 69: is "s." really an abbreviation of "see"?

guess i would rather keep "see", saving 1 char is a bit pointless. :-)

ThomasWaldmann: please add a dot at the end of the sentences.

multigeiger/userdefines-example.h
```diff
@@ -63,3 +63,20 @@
...
+// Accumulated dose rate threshold to play the local alarm
+// The accumulated dose rate is the overall average since the last start of MultiGeiger
+// Default value: 0.500 µSv/h
+#define LOCAL_ALARM_THRESHOLD 0.500  // µSv/h
+
+// Factor of current dose rate vs. accumulated dose rate to play the local alarm
+// Default value: 3
+// Example: accumulated dose rate from days of operation is at 0.1 µSv/h
+// If current dose rate rises to (default) 3 times the accumulated dose rate, i.e. 0.3 µSv/h, sound the alarm
+#define LOCAL_ALARM_FACTOR 3  // current / accumulated dose rate
```
ThomasWaldmann, line 82: can we say "trigger the alarm" instead of "play" / "sound" alarm?

i plan to add a LED to signal misc stuff later, so there might be more alarm options due to that.

multigeiger/userdefines-example.h
```diff
@@ -63,3 +63,20 @@
 // 0x2A38: Heart Rate Sensor Position --> sends TUBE_TYPE
 // 0x2A39: Heart Rate Control Point --> allows to reset "energy expenditure", as required by service definition
 #define SEND2BLE false
+
+// Play an alarm sound when radiation level is too high?
+// Activates when either accumulated dose rate reaches the set threshold (s. below)
+// or when the current dose rate is higher than the accumulated dose rate by the set factor (s. below)
+// ! Requires a valid tube type to be set in order to calculate dose rate
+#define LOCAL_ALARM false
```
ThomasWaldmann, line 71: ALARM_SOUND or AUDIO_ALARM maybe?

multigeiger/speaker.cpp
```diff
@@ -160,6 +169,12 @@ void tick_enable(bool enable) {
   }
 }
 
+void alarm() {
+  portENTER_CRITICAL_ISR(&mux_audio);
+  isr_audio_sequence = alarm_sequence;
+  portEXIT_CRITICAL_ISR(&mux_audio);
+}
+
```
ThomasWaldmann, line 177: this is not called from an ISR, so you have to use the calls without _ISR.

multigeiger/webconf.cpp
```diff
@@ -51,6 +57,19 @@ iotwebconf::TextParameter deveuiParam = iotwebconf::TextParameter("DEVEUI", "dev
 iotwebconf::TextParameter appeuiParam = iotwebconf::TextParameter("APPEUI", "appeui", appeui, 17);
 iotwebconf::TextParameter appkeyParam = iotwebconf::TextParameter("APPKEY", "appkey", appkey, 33);
 
+iotwebconf::ParameterGroup grpAlarm = iotwebconf::ParameterGroup("alarm", "Local Alarm Setting");
+iotwebconf::CheckboxParameter soundLocalAlarmParam = iotwebconf::CheckboxParameter("Enable local alarm", "soundLocalAlarm", soundLocalAlarm_c, CHECKBOX_LEN, soundLocalAlarm);
```
ThomasWaldmann, line 61: alarm sound / audio alarm

multigeiger/userdefines-example.h
```diff
@@ -63,3 +63,20 @@
 // 0x2A38: Heart Rate Sensor Position --> sends TUBE_TYPE
 // 0x2A39: Heart Rate Control Point --> allows to reset "energy expenditure", as required by service definition
 #define SEND2BLE false
+
+// Play an alarm sound when radiation level is too high?
+// Activates when either accumulated dose rate reaches the set threshold (s. below)
+// or when the current dose rate is higher than the accumulated dose rate by the set factor (s. below)
+// ! Requires a valid tube type to be set in order to calculate dose rate
```
ThomasWaldmann, line 70: True, but rather should be noted for 76 and 82.

ThomasWaldmann (changes requested):
lots of nitpicking, but guess the log level and the kind of mutex call might be really important.

multigeiger/multigeiger.ino
t-pi: Agreed, changed to WARNING

multigeiger/speaker.cpp
t-pi: guess it was just a bad phrasing by me, corrected

multigeiger/speaker.cpp
t-pi: my mistake, removed.

multigeiger/userdefines-example.h
t-pi: "s." is often used as abbreviation, but indeed doesn't matter here. Added dots.

multigeiger/userdefines-example.h
t-pi: Added it in all three description blocks.

multigeiger/userdefines-example.h
t-pi: changed to LOCAL_ALARM_SOUND

multigeiger/userdefines-example.h
t-pi: ok

multigeiger/webconf.cpp
t-pi: "Enable local alarm sound" for consistency

ThomasWaldmann (approved):

ThomasWaldmann:
Thanks!

PR #432, opened 2021-06-17
remove unused ntp/time related code, fixes #429

### LINKED ISSUE #429: not calling wait_ntp (yet)?
there's an implementation, but we are not calling it.

ThomasWaldmann:


ThomasWaldmann:
https://github.com/ecocurious2/MultiGeiger/pull/432/checks?check_run_id=2850353559 @rexfue this doesn't look like related to my change here - is it caused by the last iotwebconf change?

t-pi:
We just checked: IotWebConf moved from 3.0.2 to 3.1.0 in the meantime, apparently with breaking changes...
@rexfue , could you please have a look? Thanks!

PR #433, opened 2021-06-18
Iot 3.1

rexfue:
Only three very small changes were necessary. Now it compiles with iotwebconf 3.1.0 (but no longer with elder versions).
OTA tested, works.


ThomasWaldmann:
Thanks @rexfue!

PR #435, opened 2021-06-18
bugs in speaker.cpp fixed (no static var, ISR mux improved)

t-pi:
as discussed: small bugfixes and ISR blocking improvements in speaker.cpp, to separate from local alarm feature changes

### DIFF
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -45,7 +45,7 @@ void IRAM_ATTR isr_audio() {
 
   // tone and tick generation, also led blinking
   int frequency_mHz = 0, volume = 0, led = 0, duration_ms = 0;  // init avoids compiler warning
-  static bool playing_audio = false, playing_tick = false;
+  bool playing_audio = false, playing_tick = false;
 
   // fetch next tone / next led state
   portENTER_CRITICAL_ISR(&mux_audio);
@@ -64,12 +64,9 @@ void IRAM_ATTR isr_audio() {
     volume = *p++;
     led = *p++;
     duration_ms = *p++;
-    isr_sequence = p;
+    isr_sequence = duration_ms == 0 ? NULL : p;
   }
   portEXIT_CRITICAL_ISR(&mux_audio);
-
-  if (!isr_sequence)
-    return;  // nothing to do
 
   // note: by all means, **AVOID** mcpwm_set_duty() in ISR, causes floating point coprocessor troubles!
   //       when just calling mcpwm_set_duty_**type**(), it will reuse a previously set duty cycle.
@@ -119,9 +116,9 @@ void IRAM_ATTR tick(bool high) {
   // high true: "tick" -> high frequency tick and LED blink
   // high false: "tock" -> lower frequency tock, no LED
   // called from ISR!
-  portENTER_CRITICAL_ISR(&mux_audio);
 
   int *sequence;
+  sequence = NULL;
 
   // we expect speaker_tick or led_tick to change at any time,
   // thus check it here and generate different sequences:
@@ -137,9 +134,9 @@ void IRAM_ATTR tick(bool high) {
     sequence[5] = 0;
     sequence[6] = led_tick ? (high ? 0 : -1) : -1;
     sequence[7] = 0;  // END
-  } else
-    sequence = NULL;
+  }
 
+  portENTER_CRITICAL_ISR(&mux_audio);
   isr_tick_sequence = sequence;
   portEXIT_CRITICAL_ISR(&mux_audio);
 }
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -109,8 +109,7 @@ void IRAM_ATTR isr_audio() {
     if (playing_tick) {
       isr_tick_sequence = NULL;
       playing_tick = false;
-    }
-    else if (playing_audio) {
+    } else if (playing_audio) {
       isr_audio_sequence = NULL;
       playing_audio = false;
     }
--- a/multigeiger/speaker.cpp
+++ b/multigeiger/speaker.cpp
@@ -65,11 +65,11 @@ void IRAM_ATTR isr_audio() {
     led = *p++;
     duration_ms = *p++;
     isr_sequence = p;
-  }
-  portEXIT_CRITICAL_ISR(&mux_audio);
-
-  if (!playing_audio && !playing_tick)
+    portEXIT_CRITICAL_ISR(&mux_audio);
+  } else {
+    portEXIT_CRITICAL_ISR(&mux_audio);
     return;
+  }
 
   // note: by all means, **AVOID** mcpwm_set_duty() in ISR, causes floating point coprocessor troubles!
   //       when just calling mcpwm_set_duty_**type**(), it will reuse a previously set duty cycle.
@@ -124,7 +124,6 @@ void IRAM_ATTR tick(bool high) {
   // called from ISR!
 
   int *sequence;
-  sequence = NULL;
 
   // we expect speaker_tick or led_tick to change at any time,
   // thus check it here and generate different sequences:
@@ -140,7 +139,8 @@ void IRAM_ATTR tick(bool high) {
     sequence[5] = 0;
     sequence[6] = led_tick ? (high ? 0 : -1) : -1;
     sequence[7] = 0;  // END
-  }
+  } else
+    sequence = NULL;
 
   portENTER_CRITICAL_ISR(&mux_audio);
   isr_tick_sequence = sequence;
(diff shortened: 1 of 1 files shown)


t-pi:
Uh not good - do not merge! Sequence not stopping!

multigeiger/speaker.cpp
```diff
@@ -45,7 +45,7 @@ void IRAM_ATTR isr_audio() {
 
   // tone and tick generation, also led blinking
   int frequency_mHz = 0, volume = 0, led = 0, duration_ms = 0;  // init avoids compiler warning
-  static bool playing_audio = false, playing_tick = false;
+  bool playing_audio = false, playing_tick = false;
```
ThomasWaldmann, line 48: i don't think this would work. it will be set to true when playing the first tone and at the end of that interrupt service, the local variable will be lost. when it plays the next tone, it will start as false again.
line 52..60 won't be executed because isr_sequence is not NULL for the non-first tone.

multigeiger/speaker.cpp
```diff
@@ -58,17 +58,17 @@ void IRAM_ATTR isr_audio() {
...
-  if (isr_sequence) {
-    volatile int *p = isr_sequence;
+  volatile int *p = isr_sequence;
+  if (p) {
     frequency_mHz = *p++;
     volume = *p++;
     led = *p++;
     duration_ms = *p++;
-    isr_sequence = p;
+    isr_sequence = duration_ms == 0 ? NULL : p;
```
ThomasWaldmann, line 67: i think this is not needed, see line 108.

multigeiger/speaker.cpp
t-pi: this might explain why it repeats all the time any sequence

t-pi: ok, should be fixed, code runs here also with ticks / start_sound enabled now 😙🎶

multigeiger/speaker.cpp
t-pi: removed, chose a different solution by relying on playing_tick / playing audio bools.

t-pi:
git still behaves in mysterious ways to me quite often... somehow the speaker.cpp got left out during my commit rebasing...

multigeiger/speaker.cpp
```diff
@@ -137,9 +140,9 @@ void IRAM_ATTR tick(bool high) {
     sequence[5] = 0;
     sequence[6] = led_tick ? (high ? 0 : -1) : -1;
     sequence[7] = 0;  // END
-  } else
-    sequence = NULL;
+  }
```
ThomasWaldmann, line 143: if you want to do it that way, you can also just keep the assignment in the else-branch.

multigeiger/speaker.cpp
```diff
@@ -68,8 +68,8 @@ void IRAM_ATTR isr_audio() {
   }
   portEXIT_CRITICAL_ISR(&mux_audio);
 
-  if (!isr_sequence)
-    return;  // nothing to do
+  if (!playing_audio && !playing_tick)
+    return;
```
ThomasWaldmann, line 72: yeah, that should work.

ThomasWaldmann: ehrm, guess that is equivalently bad as the isr_sequence reading we had before - both are variables changed by isr code.

so better revert to the "p" method.

multigeiger/speaker.cpp
t-pi: alright, done.

multigeiger/speaker.cpp
t-pi: Then we have the issue with every solution, as also "p" is part of the isr_sequence commands. I propose now an else branch with a second EXIT mux_audio, as this seems the way to go. Any new variable will also be part of the isr_sequence, right?

ThomasWaldmann: p is a local dynamic variable, so each user has its own copy of it.
the mux is only needed if multiple users access the same memory location (global or local static variable).

t-pi: Hmm, but in order to check for NULL, we have to lift its scope up to the isr_audio function. And according to [this description](https://www.arduino.cc/reference/de/language/variables/variable-scope-qualifiers/volatile/), volatile does not really help here. 
What about the if-else proposition?

ThomasWaldmann: it will be still local and dynamically allocated within the isr_audio function, so it is neither used by other functions nor by other invocactions of isr_audio (not sure if that can happen).

that volatile description is likely meant for the 8bit arduino, so parts may not apply to esp32.

ThomasWaldmann: btw, guess the volatile is not really needed here, but also does no harm.

t-pi: No, I still don't like it. 
We get `p == NULL` when the sequence reaches the end and the routine returns, but the last sequence part is not processed and the isr*_sequences are not reset, nor are the playing_ bools. So I opt 

- either for the current proposition with repeated `portEXIT `in if-else
- or for a new `portENTER - portEXIT` block around a 'playing_tick || playing_audio' if-loop

Both process the last sequence part correctly.
What would you prefer?

t-pi: ...and the second option would also require to 'portEXIT's

t-pi:
```
portENTER_CRITICAL_ISR(&mux_audio);
if (!playing_tick && !playing_audio) {
   portEXIT_CRITICAL_ISR(&mux_audio);
   return;
}
portEXIT_CRITICAL_ISR(&mux_audio);
```

ThomasWaldmann: not sure what you mean / what the problem is.

If we assign to p within the mutually exclusive block, then no other mutually exclusive block (using the same mutex) will run at the same time. so especially the mutex block at the end will not have partially executed.

multigeiger/speaker.cpp
```diff
@@ -65,11 +65,11 @@ void IRAM_ATTR isr_audio() {
     led = *p++;
     duration_ms = *p++;
     isr_sequence = p;
```
t-pi, line 67: @ThomasWaldmann: 1) Here we defined the frequency, volume, led and duration for the last sequence. The values are valid, as isr_sequence was not zero, so their interpretation in the second part of the function should be carried out. If duration is now 0, it's the last part of the sequence and p (=isr_sequence) is NULL.

multigeiger/speaker.cpp
```diff
@@ -65,11 +65,11 @@ void IRAM_ATTR isr_audio() {
...
+    portEXIT_CRITICAL_ISR(&mux_audio);
+  } else {
+    portEXIT_CRITICAL_ISR(&mux_audio);
+    return;
   }
-  portEXIT_CRITICAL_ISR(&mux_audio);
-
-  if (!isr_sequence)
-    return;  // nothing to do
```
t-pi, line 72: 2) If now we have the last part of the sequence, isr_sequence is NULL, however the parameters (frequency, volume, led, duration) have not yet been interpreted, i.e. the sequences have not been NULLed yet. But nevertheless we return here. So p / isr_sequence are not the right criteria to return.

multigeiger/speaker.cpp
t-pi: I've commented in the original version, s. below

multigeiger/speaker.cpp
ThomasWaldmann: if will only execute 65+ if p is not NULL.
if duration is 0, it will **later** be set p to NULL, but that is inside a mutex block.

multigeiger/speaker.cpp
ThomasWaldmann: i was suggesting to use !p here (p being assigned in the mutex block right above).
p is not NULL if there is something to do still.

multigeiger/speaker.cpp
t-pi: Right. but this later will not happen for the last block, as p will be NULL then and an 'if (!p)' will return.

multigeiger/speaker.cpp
t-pi: Yes, but imo !p has the same issue, being NULL after the last sequence part has been parsed, despite the corresponding paramaters have not yet been processed.

multigeiger/speaker.cpp
```diff
@@ -140,6 +142,7 @@ void IRAM_ATTR tick(bool high) {
   } else
     sequence = NULL;
 
+  portENTER_CRITICAL_ISR(&mux_audio);
   isr_tick_sequence = sequence;
   portEXIT_CRITICAL_ISR(&mux_audio);
```
ThomasWaldmann, line 147: i thought about this:
- this was intended as an optimization (keep the mutex for a shorter time)
- sequence pointer is local, but it points to the global tick_sequence array.
- so i now think this change is incorrect: in-parallel modifications of the global array should be avoided by the mutex.
- maybe it is not actually causing a problem right now if we can assume that only one tick() call is running at a specific time, but i am not sure about how ISRs are processed on the ESP32.
- but from a structural pov it is better to have the mutex while mutating global stuff.

ThomasWaldmann: this also needs checking in the other PR / in the new code that was added, maybe another similar problem there.

t-pi:
Covered in #436, closing here

PR #441, opened 2021-09-10
fix uptime display - casting brackets

t-pi:
Casting brackets misplaced on TimeSec for displaying current uptime.
All variables for seconds changed to "unsigned"

### DIFF
--- a/multigeiger/display.cpp
+++ b/multigeiger/display.cpp
@@ -131,8 +131,8 @@ void display_status(void) {
 
 char *format_time(unsigned int secs) {
   static char result[4];
-  int mins = secs / 60;
-  int hours = secs / (60 * 60);
+  unsigned int mins = secs / 60;
+  unsigned int hours = secs / (60 * 60);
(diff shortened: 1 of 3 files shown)


ThomasWaldmann (approved):
Good catch! LGTM.

I'll squash the individual commits though because they were not cleanly separated anyway.

ThomasWaldmann:
Does this also mean that the uptime was that long that you noticed some overflow?

Guess that would be a good proof of stability.

ThomasWaldmann:
https://en.wikipedia.org/wiki/Bracket good article about parens / brackets / braces ...

t-pi:
Thanks for the merge - indeed, I am currently at an uptime of -16 days 😁
This is still based on the millis() to int casting with the unfortunate parentheses setting. 
(BTW thanks for the links, I was about to write parentheses, but that sounded somewhat heavy - and leo.org had "brackets" for "die Klammer  Pl.: die Klammern   - eckig, rund oder geschweift", thus good enough for me ;)

PR #444, opened 2022-02-06
Refactor GMC data processing

t-pi:
Refactor different functions for GMC processing into single one.
- DISPLAYREFRESH --> HEARTBEAT as base processing interval (default 10s)
- Other events snap to this heartbeat interval
- Improved handling of future additions (e.g. Telegram signaling) as separate event in single location
- Removed MINCOUNT as trigger for heartbeat.

### DIFF
--- a/multigeiger/multigeiger.ino
+++ b/multigeiger/multigeiger.ino
@@ -121,7 +121,7 @@ void local_alarmsequence(void) {
 }
 
 void process_GMC(unsigned long current_ms, unsigned long current_counts, unsigned long gm_count_timestamp, unsigned long current_hv_pulses,
-              bool have_thp, float temperature, float humidity, float pressure, int wifi_status) {
+                 bool have_thp, float temperature, float humidity, float pressure, int wifi_status) {
   struct GM_State {
     unsigned long timestamp;
     unsigned long counts;
@@ -144,7 +144,7 @@ void process_GMC(unsigned long current_ms, unsigned long current_counts, unsigne
   };
 
   // millis(), counts or hv_pulses overflow?
-  for (int i = 0; i<savedstates_count; i++) {
+  for (int i = 0; i < savedstates_count; i++) {
     if (current_ms < saved_state[i].timestamp) saved_state[i].timestamp = current_ms;
     if (current_counts < saved_state[i].counts) saved_state[i].counts = current_counts;
     if (gm_count_timestamp < saved_state[i].last_count_timestamp) saved_state[i].last_count_timestamp = gm_count_timestamp;
@@ -157,7 +157,7 @@ void process_GMC(unsigned long current_ms, unsigned long current_counts, unsigne
 
   // Startup
   if ((gm_count_timestamp == 0) && (saved_state[HEARTBEAT].last_count_timestamp == 0)) {
-      // If there were no pulses after AFTERSTART msecs after boot, clear display anyway and show 0 counts.
+    // If there were no pulses after AFTERSTART msecs after boot, clear display anyway and show 0 counts.
     static unsigned long boot_timestamp = millis();
     static unsigned long afterStartTime = AFTERSTART * 1000;
     if (afterStartTime && ((current_ms - boot_timestamp) >= afterStartTime)) {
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
diff isn't easy to review, would've been better done in multiple commits with special focus in each.

t-pi:
Sorry, was struggling a bit to get it working - and did not want to commit some kind of intermediate failing status

ThomasWaldmann:
the current code mix is not pretty (unrelated to this PR), maybe we could rather have something like this:

- keep some global global gm counts + thp historical values
- have some code that ONLY updates it with new data (not calling any other functions)
- have some code that computes measurements / averages / min / max from the historical data (could be also kept in globals)
- have some code that sends / alarm-checks measurements

we could do fancy graphical stuff if we had some arrays with gm counts, thp for the last N secs / mins / hours.

guess we'ld need some "ring buffers" to easily keep / update historic values in a limited memory.

t-pi:
Agree about the current code mix. But I guess that ship has sailed...

Regarding the variables, at the moment we keep the main counters in the main loop - and with this PR the specific historic states as needed per tracked event in the process_GMC function. at least a little bit cleaner than before :)

IMO a more general approach to keep historic data reaches beyond this PR, as several things have to be considered: values to be kept, frequency, aggregation, memory to be spent - and not least why would we do this. 

Here I was aiming to prepare a better handling of the Telegram (and other) messages which are not necessarily in sync with the sensor.community etc. updates.

t-pi:
@ThomasWaldmann would it be possible to merge this PR? I would like to rebase Telegram on it and add MQTT as well as BME680 AQI, too.
Or else, what needs to be done prior to a merge?
Thanks!

multigeiger/multigeiger.ino
```diff
@@ -268,32 +273,31 @@ void loop() {
...
-  update_ble_status();
-
-  // do any other periodic updates for uplinks
-  poll_transmission();
+  read_hv(&hv_error, &hv_pulses);
+  set_status(STATUS_HV, hv_error ? ST_HV_ERROR : ST_HV_OK);
 
-  publish(current_ms, gm_counts, gm_count_timestamp, hv_pulses, temperature, humidity, pressure);
+  read_GMC(&gm_counts, &gm_count_timestamp, &gm_count_time_between);
+  read_THP(current_ms, &have_thp, &temperature, &humidity, &pressure);
```
ThomasWaldmann, line 285: why did you change order here?

multigeiger/multigeiger.ino
```diff
@@ -106,138 +103,146 @@ int update_ble_status(void) {  // currently no error detection
...
+        log(DEBUG, "Measured GM: cpm= %d HV=%d", current_cpm, hv_pulses);
+        transmit_data(tubes[TUBE_TYPE].type, tubes[TUBE_TYPE].nbr, dt, hv_pulses, counts, current_cpm,
+                      have_thp, temperature, humidity, pressure, wifi_status);
+        break;
+      case ONE_MINUTE:
+        if (Serial_Print_Mode == Serial_One_Minute_Log)
+          log_data_one_minute((current_ms / 1000), current_cpm, counts);
+        break;
+      default:
+        break;
```
ThomasWaldmann, line 239: something missing here?

multigeiger/multigeiger.ino
```diff
@@ -106,138 +103,146 @@ int update_ble_status(void) {  // currently no error detection
...
+  };
+
+  // Events
+  const byte HEARTBEAT = 0;
+  const byte MEASUREMENT = 1;
+  const byte ONE_MINUTE = 2;
+
+  const int savedstates_count = 3;
+
+  static GM_State saved_state[savedstates_count];
```
ThomasWaldmann, line 139: are the array members in a defined state now, like e.g. is it documented that this is all-zero?

ThomasWaldmann (commented):
well, as i already said:

i can't really review this, too many changes mixed together.
i commented some stuff i noticed.

maybe someone else can or you just make sure yourself that everything is behaving and computing correctly.

PR #450, opened 2022-02-26
Update platformio-example.ini with "min_spiffs.csv" option

pjgueno:
For old Heltec WiFiKit 32 with only 4MB. 
Minimum SPIFFS added for Platformio as described for Arduino IDE, in order to extend the flashable memory.

### DIFF
--- a/platformio-example.ini
+++ b/platformio-example.ini
@@ -16,6 +16,7 @@ src_dir = multigeiger
 
 [env:geiger]
 board = heltec_wireless_stick
+board_build.partitions = min_spiffs.csv
 build_flags =
   -D CFG_eu868=1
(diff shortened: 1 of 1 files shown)


platformio-example.ini
```diff
@@ -16,6 +16,7 @@ src_dir = multigeiger
 
 [env:geiger]
 board = heltec_wireless_stick
+board_build.partitions = min_spiffs.csv
```
ThomasWaldmann, line 19: does this work for all currently supported devices?

is the result the same as with the ardiuno settings described in the docs?

pjgueno: It works with the old wifi Kit 32 which have 4 MB memory.

PR #453, opened 2022-04-07
local WLAN http-pages, influx-db integration and translation system

tom-r:
Implements 

1. 2 local http-pages on sensor, which are accessible in the local WLAN
- first page is display of actual sensor values, Wifi-quality, memory, etc
- temp./humidity&pressure are displayed if BME 2/680 sensor is connected
- second page show log-infos, which are normally visible on serial output only.
- Debug level is temporarily adjustable
Purpose is to monitor the current status of the sensor being installed outside, without the necessity to take it back in.
Sensor data can also be monitored with broken Internet connection as long as WLAN is available.
2. added an influx-db data export to the Config-page.
3. added a translation system, so that language of texts displayed on the 2 pages can be selected at compile-time. Firmware is compiled then with the selected language.
the only prerequisite is to set a build-flag '-DTRANSL_XX', where XX is the language.
see platformio-example.ini for a sample. Languages DE and EN are available already.
Auswahl_010-thumb
Auswahl_011


multigeiger/translations_de.h
```diff
@@ -0,0 +1,52 @@
+/* TR, 07.04.2022
+ add new translation strings to ALL files translations_xx.h !!!
```
ThomasWaldmann, line 2: translations have to work based on the already established workflow we already use for documentation (standard gettext / po files / transifex as nice UI for translators).

do not reinvent the wheel.

and also please do not put something major like this into another pull request.

tom-r: I did not reinvent the wheel, this is all (!) ported from the Luftdaten-Sensor project, with some improvements...
This is an ESP32, not a computer, and you really have to tweak the code for memory and size consumption. You have to move this from the heap and put it into flash memory/PROGMEM as much as possible, to get it performing at least somehow. Standard gettext & po files will not really work with an ESP32. I'm not even sure it the macros for po/gettext are available for the ESP platform ( I didn't check this, to be honest).
On the other other hand : When else would you need text translations in a web page, if not upon introduction of the pages themselves ...

ThomasWaldmann: this PR is way to big and unfocussed IMHO.

it needs to get split up, please start with a PR that ONLY implements the http pages (in English).

no translations, no influxdb.

t-pi:
@TW, imo the translation for the embedded pages can differ from the translation for the documentation.
@tom-r, there is a pull request in validation to simplify somewhat the different events (currently under transmit or publish) into a single event processing function. I propose to wait for its approval and then integrate influx connection afterwards.

tom-r:
Da gibt's noch 'ne Baustelle die mir aufgefallen ist, mit den Logleveln:
In der V1.17-0-dev existieren 2 unterschiedliche Log-Zyklen, einmal der Loglevel-Kreis mit NOLOG. CRITICAL, ERROR, WARNING, INFO und DEBUG (s. log.h), wobei NOLOG unüblicherweise = 999 gesetzt ist und dann von 4 abwärts bis DEBUG = 0; also **kleinere Zahl** --> **mehr** Info
Und weiterhin der im multigeiger.ino verwendete "Serial_Print_Mode" (s. log_data.h). Hier wird **mehr** Info bei **steigenden** Werten ausgegeben.
Das ist gegenläufig und schießt sich selbst ins Knie ...
Deswegen hab ich in diesem PR den NOLOG ... DEBUG Zyklus umgedreht ( die hinterlegten Nummern und diese 'inkludiere alle kleineren Werte'-Regel). 

Üblicherweise setzt man loglevel auf 0 und hat keine Ausgabe und je höher der Wert, desto mehr Log wird ausgegeben (so wie im Serial_Print_Mode).
Zumindest kenn ich das so aus dem Berufsleben.
Dann hast' außerdem die Möglichkeit noch ein zusätzliches Level mit mehr Infos einzuführen und stehst' net bei Null an ...

Am Ende gehört aber m.E. der Serial_Print_Mode-Zyklus aus log_data.h  in die normalen Loglevel überführt ...
(_Das_ habe ich nicht gemacht )

PR #458, opened 2022-05-18
fix hal_init error upon compile

t-pi:
modify platform.ini template to compile without hal_init error


t-pi:
...it compiles and links without error, could you please test with a LoRaWan configuration, as I currently have none. Thanks

rexfue:
Tested with LoRa - it works

pbriesch:
Please merge!

